PR #51, opened 2015-01-08
Better error message when field contains invalid chars.

mtyaka:
To make it easier to understand what's wrong when trying to create a key that contains characters not allowed in the key's fields.

### DIFF
--- a/opaque_keys/edx/locator.py
+++ b/opaque_keys/edx/locator.py
@@ -193,8 +193,10 @@ class CourseLocator(BlockLocatorBase, CourseKey):
             if version_guid:
                 version_guid = self.as_object_id(version_guid)
 
-            if not all(field is None or self.ALLOWED_ID_RE.match(field) for field in [org, course, run, branch]):
-                raise InvalidKeyError(self.__class__, [org, course, run, branch])
+            for name, value in [['org', org], ['course', course], ['run', run], ['branch', branch]]:
+                if not (value is None or self.ALLOWED_ID_RE.match(value)):
+                    raise InvalidKeyError(self.__class__,
+                                          "Special characters not allowed in field {}: '{}'".format(name, value))
 
         super(CourseLocator, self).__init__(
             org=org,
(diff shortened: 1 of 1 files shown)


edx-webhook:
Thanks for the pull request, @mtyaka! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request. 

To automatically create an OSPR issue for this pull request, just visit this link: http://openedx-webhooks.herokuapp.com/github/process_pr?number=51&repo=edx%2Fopaque-keys

Kelketek:
:+1:

mtyaka:
@dmitchell Can you take a look at this, please?

mtyaka:
@cpennington Can you review this, please?

opaque_keys/edx/locator.py
```diff
@@ -193,8 +193,9 @@ def __init__(self, org=None, course=None, run=None, branch=None, version_guid=No
             if version_guid:
                 version_guid = self.as_object_id(version_guid)
 
-            if not all(field is None or self.ALLOWED_ID_RE.match(field) for field in [org, course, run, branch]):
-                raise InvalidKeyError(self.__class__, [org, course, run, branch])
+            for field in [org, course, run, branch]:
+                if not (field is None or self.ALLOWED_ID_RE.match(field)):
+                    raise InvalidKeyError(self.__class__, "Special characters not allowed in '{}'".format(field))
```
cpennington: Optional: It might be good to include the name of the field that the invalid string is passed in.

cpennington:
:+1:

opaque_keys/edx/locator.py
mtyaka: Added field name to the error message, thanks @cpennington

PR #52, opened 2015-02-26
Added Message to Exception

clintonb:
This will help debug an issue currently on stage, and perhaps help resolve future issues more quickly.

@cpennington @brianhw @doctoryes @mulby


doctoryes:
:+1:  - assuming that it'll be clear what type `other` is from its formatted representation.

clintonb:
Updated error message to list both argument types.

PR #54, opened 2015-04-29
Allow comparison of OpaqueKeys that have the same set of fields (and the...

cpennington:
... same deprecation status)

### DIFF
--- a/opaque_keys/__init__.py
+++ b/opaque_keys/__init__.py
@@ -373,8 +373,8 @@ class OpaqueKey(object):
         return not self == other
 
     def __lt__(self, other):
-        if type(self) != type(other):
-            raise TypeError('Cannot compare {} to {}.'.format(type(self), type(other)))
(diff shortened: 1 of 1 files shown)


cpennington:
@doctoryes: Need this for https://github.com/edx/edx-platform/pull/7789

opaque_keys/__init__.py
```diff
@@ -373,8 +373,8 @@ def __ne__(self, other):
         return not self == other
 
     def __lt__(self, other):
-        if type(self) != type(other):
-            raise TypeError('Cannot compare {} to {}.'.format(type(self), type(other)))
+        if (self.KEY_FIELDS, self.CANONICAL_NAMESPACE, self.deprecated) != (other.KEY_FIELDS, other.CANONICAL_NAMESPACE, other.deprecated):
+            raise TypeError("{!r} is incompatible with {!r}".format(self, other))
```
doctoryes: Looks OK to me - but which OpaqueKey derivatives are you trying to compare that need this change?

cpennington: `SlashSeparatedCourseKey` (which is deprecated, but still appears in places), and `CourseLocator`.

doctoryes:
:+1:

PR #55, opened 2015-05-05
Regex does not match asset url when it contains spaces

awaisdar001:
In course when user uploads an asset which has percentage`%` in their filenames; it faces issue on  `course re-run`.
[opaque_keys/edx/locator.py#L98](https://github.com/edx/opaque-keys/blob/master/opaque_keys/edx/locator.py#L98)
Above `Regex` parses asset names and when an asset has `%` in its `name` this Regex  fails to compile it. *(Split to Split)*

Updated the `Regex` to handle the case where if there is a `%` in asset name, it can parse successfully.

 [TNL2098](https://openedx.atlassian.net/browse/TNL-2098)

### DIFF
--- a/opaque_keys/edx/tests/test_asset_locators.py
+++ b/opaque_keys/edx/tests/test_asset_locators.py
@@ -15,6 +15,7 @@ class TestAssetLocators(TestCase):
     """
     Tests of :class:`AssetLocator`
     """
+
     @ddt.data(
         "/c4x/org/course/asset/path",
     )
@@ -48,7 +49,10 @@ class TestAssetLocators(TestCase):
     def test_deprecated_son(self, key_cls, prefix, tag, source):
         source_key = key_cls(*source, deprecated=True)
         son = source_key.to_deprecated_son(prefix=prefix, tag=tag)
-        self.assertEquals(son.keys(), [prefix + key for key in ('tag', 'org', 'course', 'category', 'name', 'revision')])
+        self.assertEquals(
+            son.keys(),
+            [prefix + key for key in ('tag', 'org', 'course', 'category', 'name', 'revision')]
+        )
 
         self.assertEquals(son[prefix + 'tag'], tag)
         self.assertEquals(son[prefix + 'category'], source_key.block_type)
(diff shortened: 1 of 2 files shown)


awaisdar001:
@adampalay @cpennington Please review.

cpennington:
Can you add a test that specifies the asset locator that you expect to be parsed? I'm a bit confused as to whether you want to support filenames (and thus block_ids) that have spaces `"foo bar"`, or ones that have encoded spaces `"foo%20bar"`.

awaisdar001:
@cpennington I have added a test in `edx-platform` in which I have imported a course which has an asset with `foo%20bar` filename which actually hit the Regex. 
For example when some uploads video transcripts himself from his system, the filename will be play its part.In this case the filename had `%` in it, hence the Regex was failing. 

Problem occurs when we re-run a split course

cpennington:
Tests in edx-platform are insufficient. This repo is meant to be standalone, and should be tested as such.

Regarding `'%'` vs `' '`: My concern is that seeing a `%` in the `block_id` might be an indication that url encoding isn't being handled properly, rather than something where we should extend opaque keys.

awaisdar001:
@cpennington If we decode the asset name, then it would convert `%20` to `' '` right? That also does not match the Regex (I mean if a filename has spaces, it also breaks). 

The problem why we can not do decode the url is, actual asset is placed on the sever with `%20` in its filename.
Changing 
`(/src/foo/bar%20bar2.srt)` 
to something like
`(/src/foo/bar bar2.srt)` 
we will unable to locate the file.

We will not be able to copy the asset to the destination course if its name is modified.

cpennington:
Ok. I'm fine with using `%20` in the block id if that's required. I just wanted to make sure we were solving the right problem.

awaisdar001:
@cpennington Yes, Do we need to write the tests for this in `Opaque-keys` repo?

awaisdar001:
@cpennington Please Review. 
I have added a test that targets that particular Regex.

adampalay:
@awaisdar001 Cale's out of the office for the rest of the week

awaisdar001:
@doctoryes Can you please review this PR?
Thanks.

opaque_keys/edx/tests/test_asset_locators.py
```diff
@@ -93,3 +93,13 @@ def test_empty_path(self):
             '/c4x/org/course/asset/',
             unicode(CourseKey.from_string('org/course/run').make_asset_key('asset', ''))
         )
+
+    @ddt.data(
+        "asset-v1:UOG+cs_34+Cs128+type@asset+block@subs_Introduction%20To%20New.srt.sjson",
+    )
```
doctoryes: You should use this ddt test to test *all* special characters, like ~, :, -, etc.

awaisdar001:
@doctoryes Comments addressed regarding checking other special characters like `~`, `:`, `-`, '.'

doctoryes:
Thanks! :+1:

adampalay:
:+1: 
@awaisdar001 , don't forget to open a PR against master to update this library

PR #56, opened 2015-06-19
 Fixed Broken page in "Schedule & Details" due to course image

awaisdar001:
`Schedule & Details` page of the course was breaking because it had empty `course_image`. 
with the empty image path, the assert locator throws `InvalidKeyError`. 

Mongo & Split were behaving differently, when a asset path is empty. Split raise `InvalidKeyError` but Mongo does not. 

[TNL-2435](https://openedx.atlassian.net/browse/TNL-2435)

### DIFF
--- a/opaque_keys/edx/tests/test_asset_locators.py
+++ b/opaque_keys/edx/tests/test_asset_locators.py
@@ -6,7 +6,6 @@ import ddt
 from unittest import TestCase
 from opaque_keys import InvalidKeyError
 from opaque_keys.edx.keys import AssetKey, CourseKey
-
 from opaque_keys.edx.locator import AssetLocator, CourseLocator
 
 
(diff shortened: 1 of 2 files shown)


awaisdar001:
@adampalay For instance, Lets think we are allowing empty paths for assets, this how it would look like.
We can have calen's views on that as well.

adampalay:
I think this is the right way to go because I suspect we have a bunch of courses with an empty path to the course image

awaisdar001:
@cpennington Please review.

cpennington:
I disagree. I think that an `AssetKey` with no image path isn't a meaningful/useful object. I'd rather have edx-platform change to handle the error (and fall back to the default image) than have `OpaqueKeys` accept an empty string and return an object that doesn't make sense.

awaisdar001:
@cpennington To my understanding, you mean, we need to throw `InvalidKeyError` for both `Mongo` and `Split` by  disallowing empty paths for `Mongo`, Right?

And the change would be in `opaque-keys` repo?
Please confirm.

cpennington:
@awaisdar001 yes, that's right.

awaisdar001:
@cpennington Comments addressed. Have a look.

opaque_keys/edx/tests/test_asset_locators.py
```diff
@@ -91,12 +90,8 @@ def test_replace(self):
...
     def test_empty_path(self):
         with self.assertRaises(InvalidKeyError):
-            CourseKey.from_string('course-v1:org+course+run').make_asset_key('asset', '')
-
-        self.assertEquals(
-            '/c4x/org/course/asset/',
-            unicode(CourseKey.from_string('org/course/run').make_asset_key('asset', ''))
-        )
+            CourseKey.from_string('course-locator:org+course+run').make_asset_key('asset', '')
+            CourseKey.from_string('org/course/run').make_asset_key('asset', '')
```
cpennington: You should call `assertRaises` around each of these calls, rather than having only one assertion. As it stands, the second line (94) will never be tested.

awaisdar001:
@cpennington Added `self.assertRaises` for both statements. Please review

cpennington:
:+1:

awaisdar001:
@symbolist @waheedahmed Another review please.

waheedahmed:
:+1:

PR #57, opened 2015-11-17
Catch UnicodeEncodeError.

HassanJaveed84:
@brianhw is this what you had in mind ?
I'm looking to add a test for this, incase of UnicodeDecodeError, should we return nothing or reraise ?

### DIFF
--- a/opaque_keys/__init__.py
+++ b/opaque_keys/__init__.py
@@ -233,7 +233,7 @@ class OpaqueKey(object):
             # Cache that the namespace doesn't correspond to a known plugin,
             # so that we don't waste time checking every time we hit
             # a particular unknown namespace (like i4x)
-            raise InvalidKeyError(cls, '{}:*'.format(namespace))
+            raise InvalidKeyError(cls, u'{}:*'.format(namespace))
 
     LOADED_DRIVERS = defaultdict()  # If you change default, change test_default_deprecated
 
--- a/opaque_keys/tests/test_opaque_keys.py
+++ b/opaque_keys/tests/test_opaque_keys.py
@@ -137,6 +137,9 @@ class KeyTests(TestCase):
         with self.assertRaises(InvalidKeyError):
             DummyKey.from_string('dict:abcd')
 
+        with self.assertRaises(InvalidKeyError):
+            DummyKey.from_string(u'\xfb:abcd')
+
     def test_unknown_namespace(self):
         with self.assertRaises(InvalidKeyError):
             DummyKey.from_string('no_namespace:0x10')


brianhw:
I think the fix is even simpler.  It should be enough to put a u on the front of the string being formatted.  So:  raise InvalidKeyError(cls, u'{}:*'.format(namespace))

brianhw:
Also, I would add a test (back in) that actually tests the fix.  Break it first by adding a test that includes non-ascii characters in the namespace (i.e. the part that precedes the colon).  Then confirm that the fix results in the test passing.

HassanJaveed84:
@brianhw when I add a test such as
``` 
    DummyKey.from_string('\xc3:abcd')
```
it throws the UnicodeDecodeError exception at ```namespace, __, rest = serialized.partition(cls.NAMESPACE_SEPARATOR)```

brianhw:
Hmm.  Good thing we're testing it!  But maybe that's not quite the right test?  I'm not sure what kind of string we need to pass in to produce the original error, but that must not be it.  Should it be a unicode string with a non-ascii character?  I don't know what \xc3 is, but not utf-8, and the whole thing is a str, not unicode, so I'm guessing a UnicodeDecodeError might be a logical outcome.

HassanJaveed84:
From one of our other conversations over the email, you shared a stacktrace
raise InvalidKeyError(cls, \'{}:*\'.format(namespace))\nUnicodeEncodeError: \'ascii\' codec can\'t encode character u\'\\xfb\' in position 7: ordinal not in range(128)\n\n

brianhw:
I don't know what I sent you, but here's another example from job/analyze-event-dev/13/console: 
```
=>    course_key = CourseKey.from_string(course_id_string) 
=>    return cls.get_namespace_plugin(namespace)._from_string(rest)
=>    raise InvalidKeyError(cls, \'{}:*\'.format(namespace))

UnicodeEncodeError: \'ascii\' codec can\'t encode character u\'\\u200b\' in position 8: ordinal not in range(128)\n'
```
So it looks like using a namespace that contains u'\xfb' or u'\u200b' should trigger the error.  

Is that what you meant?  Sorry if I'm not understanding your point with the stacktrace.

HassanJaveed84:
@brianhw updated.

brianhw:
LGTM.  :+1:   I would like to get a second thumb, however.

brianhw:
@cpennington Please take a quick look to confirm this change.

brianhw:
Verbal thumbs from Cale. :+1:

cpennington:
:+1:

PR #58, opened 2015-11-24
Fix UnicodeEncodeError.

HassanJaveed84:
@brianhw


opaque_keys/edx/tests/test_course_locators.py
```diff
@@ -44,6 +44,10 @@ def test_invalid_format_location(self, course_id):
         with self.assertRaises(InvalidKeyError):
             CourseLocator.from_string(course_id)
 
+    def test_invalid_org_id(self):
+        with self.assertRaises(InvalidKeyError):
+            CourseLocator(org=u'\ufffd', course='course', run='run')
+
```
brianhw: I would instead add this to the ddt input to test_course_constructor_bad_package_id() below.  It will run the same test, but will test org, course, and run.  This will also uncover a need for a u before a format string within the test, in the fourth clause:  CourseKey.from_string(u'course-v1:test+{}+2014_T2'.format(bad_id)).  That looks like it should cover things.

brianhw:
LGTM. :+1:     FYI @cpennington  we found another case.

feanil:
@brianhw this got merged but edx-platform never got updated to use this new version of opaque keys.  It also looks like a new release of opaque keys was not done.

PR #60, opened 2016-01-19
WIP: Cache OpaqueKey._key calculation.

ormsbee:
The `._key` property is used internally for equality and hashing operations, and is recalculated every time. While it's not a very expensive operation, it happens frequently enough where we see a small speedup in some test cases (~3-4%) from caching this operation.

### DIFF
--- a/opaque_keys/__init__.py
+++ b/opaque_keys/__init__.py
@@ -92,7 +92,7 @@ class OpaqueKey(object):
     Deserialization is performed by the :meth:`from_string` method.
     """
     __metaclass__ = OpaqueKeyMetaclass
-    __slots__ = ('_initialized', 'deprecated')
+    __slots__ = ('_initialized', 'deprecated', '_lazy_key')
 
     KEY_FIELDS = []
     CANONICAL_NAMESPACE = None
@@ -268,6 +268,7 @@ class OpaqueKey(object):
         # a flag used to indicate that this instance was deserialized from the
         # deprecated form and should serialize to the deprecated form
         self.deprecated = kwargs.pop('deprecated', False)
+        self._lazy_key = None
 
         if self.CHECKED_INIT:
             self._checked_init(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


ormsbee:
Okay, so stupid question: In what situations would an equality comparison of `OpaqueKey._key` differ from an equality comparison of their `unicode()` serializations? I ask because there are a number of places where we do a non-trivial amount of work to take a bunch of `course_id` strings and deserialize them, only to do equality comparisons. If the results wouldn't differ, we could just skip the deserialization entirely and do much cheaper string compares instead.

opaque_keys/__init__.py
```diff
@@ -364,7 +366,12 @@ def __getstate__(self):
     @property
     def _key(self):
         """Returns a tuple of key fields"""
-        return tuple(getattr(self, field) for field in self.KEY_FIELDS) + (self.CANONICAL_NAMESPACE, self.deprecated)  # pylint: disable=no-member
+        if self._lazy_key is None:
+            self._lazy_key = (
+                tuple(getattr(self, field) for field in self.KEY_FIELDS) +
+                (self.CANONICAL_NAMESPACE, self.deprecated)  # pylint: disable=no-member
+            )
+        return self._lazy_key
```
cpennington: Maybe just use `@lazy` instead of `@property` here?

cpennington:
I'm not immediately aware of any places where serialized equality is different than deserialized equality, and given that we use the serialized version as a database key, I think it would be safe to say that we want to have serialized equality <=> deserialized equality.

I'm not sure if we have a set of semantic tests for any opaque key implementation to satisfy yet (the opaque key laws, so to speak), but we probably should.

doctoryes:
Agreed on the equality of serialized and deserialized equality - and agreed that that property should be tested.

If we were making a 30 FPS game, I'd also recommend we compare string hashes instead of strings - but I doubt that's where we're spending our budget compared to other things. :)

ormsbee:
Okay, so I think I'm going to rework this so that it just lazily computes the serialization (or does a simple copy at initialization if it's a `from_string()`) and uses the string compares for comparison and hashing.

FWIW, this isn't actual Performance team work this sprint, just me playing with test optimization over the weekend. Doing some experiment work in edx-platform#11253, which will probably lead to a few separate PRs.

PR #62, opened 2016-01-25
Deploying to PyPI/Version Update

### LINKED ISSUE #59: Release to PyPI
* Create a new version
* Update Travis to automatically [release new tags to PyPI](https://docs.travis-ci.com/user/deployment/pypi/) (this may require renaming the package to edx-opaque-keys to avoid confusion)

clintonb:
Fixes #59

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,4 +61,3 @@ coverage.xml
 
 # Sphinx documentation
 docs/_build/
-
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,3 +11,12 @@ script:
   - make quality
 after_success:
   - coveralls
+deploy:
+  provider: pypi
+  user: edx
+  password:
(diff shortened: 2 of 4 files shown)


clintonb:
FYI @cpennington. I will also create a new tag so that can cleanup external references that continue to use various git hashes.

cpennington:
Ok. Are you also going to push to PyPI?

clintonb:
I would like to, and created #59 to do so. However, the last question about naming is what has stopped me. Should we continue calling this package `opaque-keys` or use a more transparent name, `edx-opaque-keys` that makes it clear this is intended for use with edX code?

I can go either way. A rename should not pose much developer pain. Anyone wanting to use the PyPI version would need to update requirements.txt anyway. They'd just use the new name instead of the old one.

(Puns intended!)

cpennington:
I'd be fine with clarifying the package name by prefixing edx.

clintonb:
Good...because it turns out pip doesn't like `setup.py`'s `dependency_links`, so we will need to deploy to PyPI.

PR #64, opened 2016-01-25
Cleanup

clintonb:
Bringing this repo up to standard with our other library repos.


pylintrc_tweaks
```diff
@@ -0,0 +1,5 @@
+[MASTER]
+load-plugins = edx_lint.pylint
+
+[MESSAGES CONTROL]
+DISABLE+= ,literal-used-as-attribute
```
nedbat: Why disable this?

clintonb: I am not too familiar with this code base. The calls where this error occurred seem legitimate, so I opted not to change them.

nedbat: I see this:
```
            if hasattr(cls, 'deprecated_fallback'):
                return getattr(cls, 'deprecated_fallback')._from_deprecated_string(serialized)
```
which should be:
```
    return cls.deprecated_fallback._from_deprecated_string(serialized)
```
and
```
        son[prefix + 'revision'] = getattr(self.course_key, 'branch')
```
which should be:
```
        son[prefix + 'revision'] = self.course_key.branch
```

nedbat:
@clintonb I'm always happy to help with tooling questions if need be.  I notice this was merged with no thumbs, I can help next time.

PR #65, opened 2016-01-27
Clean literal-used-as-attribute Pylint violations

clintonb:
This commit actually fixes the violations rather than ignore them.


clintonb:
@nedbat please review

nedbat:
nice :) :+1:

clintonb:
Commits squashed. Reverted change to setup.py as #66 will need to be merged before another release. Will merge this PR after tests finish.

PR #66, opened 2016-01-27
Including opaque_keys.edx.tests in exported packages

clintonb:
This package is needed by ccx-keys.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,14 +1,11 @@
-from setuptools import setup
+from setuptools import setup, find_packages
 
 setup(
     name="edx-opaque-keys",
-    version="0.2.0",
+    version="0.2.1",
     author="edX",
     url="https://github.com/edx/opaque-keys",
-    packages=[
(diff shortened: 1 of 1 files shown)


clintonb:
FYI @feanil @nedbat @benpatterson

setup.py
```diff
@@ -2,12 +2,13 @@
...
 setup(
     name="edx-opaque-keys",
-    version="0.2.0",
+    version="0.2.1",
     author="edX",
     url="https://github.com/edx/opaque-keys",
     packages=[
         "opaque_keys",
         "opaque_keys.edx",
+        "opaque_keys.edx.tests",
```
nedbat: Perhaps it would be safest now to include everything.  "opaque_keys.tests" is still missing.

clintonb: Good point. Will do.

clintonb:
@nedbat updated

PR #67, opened 2016-02-19
PERF: Optimize key hashing and equality comparison

ormsbee:
Okay folks, thank you for your patience. I think this is now in a reviewable state.

Looking at the progress page of a fairly large course (~1.3K blocks) here are three things that happen over and over to OpaqueKeys 

1. Hashing (114K times)
2. Serialization to unicode (49K times)
3. Equality checks (19K times)

This PR tries to address those issues by:

1. Storing the hash explicitly as an integer attribute, so we can throw it back with no work.
2. Storing the serialized unicode representation (which is simple, since that's usually how it's constructed).
3. Intern the OpaqueKeys.

The third one helps avoiding the extra work of hashing and serialization, but it also makes equality faster because we can short circuit them using object identity and hash value (both of which are cheap to check).

My biggest concern is the possible memory leak in the explicit cache pool. I use a `WeakValueDictionary`, but I don't know as that's enough. On the bright side, I think we create far fewer OpaqueKeys during a given request now...

@nedbat, @nasthagiri, @doctoryes, @cpennington: May I get a review from two of you folks?

### DIFF
--- a/opaque_keys/__init__.py
+++ b/opaque_keys/__init__.py
@@ -9,6 +9,7 @@ formats, and allowing new serialization formats to be installed transparently.
 from _collections import defaultdict
 from abc import ABCMeta, abstractmethod
 from functools import total_ordering
+from weakref import WeakValueDictionary
 
 from six import (
     iteritems,
@@ -98,7 +99,17 @@ class OpaqueKey(with_metaclass(OpaqueKeyMetaclass)):
     Serialization of an :class:`OpaqueKey` is performed by using the :func:`unicode` builtin.
     Deserialization is performed by the :meth:`from_string` method.
     """
-    __slots__ = ('_initialized', 'deprecated')
+    __slots__ = (
+        '_initialized',
+        'deprecated',
+
+        # Performance related
+        '_unicode',  # Cached Unicode representation
+        '_hash',  # Cache of the hash() so we don't have to recompute it so often.
+        '_cached_key',  # Cache _key representation, useful for repeated equality checks.
+        '__weakref__'  # To allow us to use the _cache_pool
+    )
+    _cache_pool = WeakValueDictionary()
 
     KEY_FIELDS = []
     CANONICAL_NAMESPACE = None
@@ -168,10 +179,19 @@ class OpaqueKey(with_metaclass(OpaqueKeyMetaclass)):
         """
         Serialize this :class:`OpaqueKey`, in the form ``<CANONICAL_NAMESPACE>:<value of _to_string>``.
         """
+        # OpaqueKeys are often repeatedly serialized, so we cache this value.
+        if self._unicode is not None:
+            return self._unicode
+
+        # TODO Revisit assigning-non-slot comments on pylint upgrade
         if self.deprecated:
             # no namespace on deprecated
-            return self._to_deprecated_string()
-        return self.NAMESPACE_SEPARATOR.join([self.CANONICAL_NAMESPACE, self._to_string()])  # pylint: disable=no-member
+            self._unicode = self._to_deprecated_string()  # pylint: disable=assigning-non-slot
+        else:
+            self._unicode = self.NAMESPACE_SEPARATOR.join(  # pylint: disable=assigning-non-slot
+                [self.CANONICAL_NAMESPACE, self._to_string()]  # pylint: disable=no-member
+            )
+        return self._unicode
 
     @classmethod
     def from_string(cls, serialized):
@@ -183,6 +203,17 @@ class OpaqueKey(with_metaclass(OpaqueKeyMetaclass)):
         Args:
             serialized: A stringified form of a :class:`OpaqueKey`
         """
+        # OpaqueKeys are immutable, so we can share them. Sharing the keys gives
+        # three benefits:
+        #    1. It reduces parsing/construction costs for duplicate OpaqueKeys.
+        #    2. It allows us to reduce the number of times we do the computation
+        #       for unicode() and hash() -- these are cached on the individual
+        #       OpaqueKey.
+        #    3. Equality checks between two OpaqueKeys are very cheap if they're
+        #       pointing at the same object (which is a frequent occurrence).
+        if serialized in cls._cache_pool:
+            return cls._cache_pool[serialized]
+
         if serialized is None:
             raise InvalidKeyError(cls, serialized)
 
@@ -191,10 +222,16 @@ class OpaqueKey(with_metaclass(OpaqueKeyMetaclass)):
         cls._drivers()
         try:
             namespace, rest = cls._separate_namespace(serialized)
-            return cls.get_namespace_plugin(namespace)._from_string(rest)
+            key = cls.get_namespace_plugin(namespace)._from_string(rest)
+            key._unicode = serialized
+            cls._cache_pool[serialized] = key
+            return key
         except InvalidKeyError:
             if hasattr(cls, 'deprecated_fallback'):
-                return cls.deprecated_fallback._from_deprecated_string(serialized)
+                key = cls.deprecated_fallback._from_deprecated_string(serialized)
+                key._unicode = serialized
+                cls._cache_pool[serialized] = key
+                return key
             raise InvalidKeyError(cls, serialized)
 
     @classmethod
@@ -271,6 +308,10 @@ class OpaqueKey(with_metaclass(OpaqueKeyMetaclass)):
         # The __init__ expects child classes to implement KEY_FIELDS
         # pylint: disable=no-member
 
+        # TODO Revisit assigning-non-slot comments on pylint upgrade
+        self._unicode = None  # pylint: disable=assigning-non-slot
+        self._cached_key = None  # pylint: disable=assigning-non-slot
+
         # a flag used to indicate that this instance was deserialized from the
         # deprecated form and should serialize to the deprecated form
         self.deprecated = kwargs.pop('deprecated', False)  # pylint: disable=assigning-non-slot
@@ -279,6 +320,7 @@ class OpaqueKey(with_metaclass(OpaqueKeyMetaclass)):
             self._checked_init(*args, **kwargs)
         else:
             self._unchecked_init(**kwargs)
+
         self._initialized = True  # pylint: disable=assigning-non-slot
 
     def _checked_init(self, *args, **kwargs):
@@ -327,10 +369,11 @@ class OpaqueKey(with_metaclass(OpaqueKeyMetaclass)):
             return self
 
         existing_values.update(kwargs)
+
         return type(self)(**existing_values)
 
     def __setattr__(self, name, value):
-        if getattr(self, '_initialized', False):
+        if name != '_unicode' and name != '_hash' and name != '_cached_key' and getattr(self, '_initialized', False):
             raise AttributeError("Can't set {!r}. OpaqueKeys are immutable.".format(name))
 
         super(OpaqueKey, self).__setattr__(name, value)  # pylint: disable=no-member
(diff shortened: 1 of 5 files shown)


ormsbee:
FWIW, looking at this again and discussing it with @cpennington, I'm going to see if we can store the explicit hash int as I believe @doctoryes suggested earlier, and check how that affects progress page profiling numbers.

ormsbee:
@efischer19: Just a quick heads up: I made a minor update to this that should make the hashing faster. Want to experiment just a little more, and then I'll look at fixing the test failures.

Thank you.

efischer19:
@ormsbee Cool, sounds great. Let me know if I can assist in any way!

opaque_keys/__init__.py
```diff
@@ -362,11 +392,13 @@ def __getstate__(self):
...
     def _key(self):
         """Returns a tuple of key fields"""
+        return unicode(self)
         # pylint: disable=no-member
-        return tuple(getattr(self, field) for field in self.KEY_FIELDS) + (self.CANONICAL_NAMESPACE, self.deprecated)
+        # return tuple(getattr(self, field) for field in self.KEY_FIELDS) + (self.CANONICAL_NAMESPACE, self.deprecated)
 
     def __eq__(self, other):
-        return isinstance(other, OpaqueKey) and self._key == other._key  # pylint: disable=protected-access
+        return unicode(self) == unicode(other)
```
doctoryes: Why not use `return hash(self) == hash(other)` here? This code still does string comparison.

ormsbee: I was a little worried about correctness and collisions, given that `hash()` is going to spit out a 32-bit integer. That might be overly paranoid, considering how tightly opaque keys are constrained in terms of format.

Come to think on it, I'm still a little worried about the correctness now, since with the current code I have, unicode strings and the opaque keys they parse into will evaluate as equal, and that's some JavaScript-like madness right there. :-P

Also, I've just made an object pool for these things, so maybe we can make it even faster by just doing an `is` comparison at the top. I'll play with a couple different options and post the results.

cpennington: Count me in as a :-1: on `key == unicode(key)`.

ormsbee: @cpennington: Yeah, that was clearly a mistake on my part.

ormsbee:
Looked at this more last night. The main issue I'm running into now is correctness on equality comparisons, since we seem to have situations where non-equal blocks serialize into the same string. There's also a bit of ambiguity whether the serialized thing we're holding onto is a string or unicode object. Need to dig more.

nasthagiri:
> There's also a bit of ambiguity whether the serialized thing we're holding onto is a string or unicode object.

@ormsbee I'm so glad you're tackling this.  This may address the odd behavior I was finding (as I mentioned earlier) when debugging BlockTransformer related work - where opaque key comparison issues prevented successful cache lookups.

ormsbee:
(sigh) There's still something fishy about the performance profile of this. Sorry, nvm on the review request yet. As you folks probably noticed, I'm doing this work in the late evenings, so I'll update tonight.

nasthagiri:
Thanks @ormsbee !

ormsbee:
@nedbat, @doctoryes: Ready for review! Looking at about a 3X improvement for a large course on my devstack for CPU, though I/O may be more dominant in our prod envs. :)

ormsbee:
While this work was mostly geared towards speeding up the progress page, it looks like the courseware index page gets a 5-10% boost as a side-effect of this change as well.

ormsbee:
And about 12-13% off of the total test execution time. <sigh> Should have done this a year ago.

nasthagiri:
That's great!

In all of our profiling work since last year, these low-level opaque-key operations have always taken a considerable amount of cumulative time. I'm so glad you've taken this on!

efischer19:
:fireworks: :fireworks: :fireworks:

ormsbee:
I have a couple small tests to write around caching behavior and I want to
manually verify this doesn't blow up on old mongo, CCX, or libraries. I'll
squeeze that in sometime this weekend.

opaque_keys/__init__.py
```diff
@@ -327,10 +371,11 @@ def replace(self, **kwargs):
             return self
 
         existing_values.update(kwargs)
+
         return type(self)(**existing_values)
 
     def __setattr__(self, name, value):
-        if getattr(self, '_initialized', False):
+        if name != '_unicode' and name != '_hash' and name != '_cached_key' and getattr(self, '_initialized', False):
```
nedbat: I wonder if a set membership check would be faster than the three comparisons.

opaque_keys/__init__.py
```diff
@@ -357,6 +402,8 @@ def __setstate__(self, state_dict):
             if key in self.KEY_FIELDS:  # pylint: disable=no-member
                 setattr(self, key, state_dict[key])
         self.deprecated = state_dict['deprecated']  # pylint: disable=assigning-non-slot
+        self._unicode = None  # pylint: disable=assigning-non-slot
```
nedbat: Why do we need all the `disable=assigning-non-slot` pragmas?  `self._unicode` is in `__slots__`.

opaque_keys/__init__.py
ormsbee: Good point. I'll try it out. Though honestly, the `__setattr__` isn't called very often relative to the others, since there are three cachable fields, and < 5K keys generated now, even on large courses.

opaque_keys/__init__.py
ormsbee: I honestly don't know. I thought this might fall in the category of "pylint gets a little wonky when you start doing evil things". I'll check again, but I'm pretty sure I added this in direct response to a pylint error message.

opaque_keys/__init__.py
```diff
@@ -183,6 +208,16 @@ def from_string(cls, serialized):
         Args:
             serialized: A stringified form of a :class:`OpaqueKey`
         """
+        # OpaqueKeys are immutable, so we can share them. Sharing the keys gives
+        # two benefits:
```
doctoryes: Three benefits below.

opaque_keys/__init__.py
```diff
@@ -183,6 +208,16 @@ def from_string(cls, serialized):
...
+        #    1. Remove parsing/construction costs.
+        #    2. Remove the costs to calculate the hash() and to_unicode() which
+        #       are cached on the individual OpaqueKey
+        #    3. It short circuits True if they're pointing at the same object,
+        #       which is a cheap comparison (and frequent).
```
doctoryes: A re-wording might help this comment, like:
`Inexpensive comparsion if pointing at the same object (which is a frequent occurrence).`

opaque_keys/edx/locator.py
```diff
@@ -260,7 +265,31 @@ def _from_string(cls, serialized):
...
+        # have to use parsed_values as a lookup key. The regex parsing of values
+        # is only about 1/4th as expensive as CourseLocator construction, and
+        # that doesn't count the much higher cost of hash and eq calculations
+        # that are avoided by using a shared CourseLocator object.
+        #
+        # Because __parsed_fields_to_keys gets mangled into something local to
+        # this class, we don't have to worry about collisions with our
+        # subclasses.
+        if parsed_values in cls.__parsed_fields_to_keys:
+            return cls.__parsed_fields_to_keys[parsed_values]
```
doctoryes: Where is this cache filled? I don't see a `set` anywhere for the cache.

ormsbee: Ugh. Refactor gone wrong. Thank you.

opaque_keys/__init__.py
ormsbee: Yeah, getting rid of that causes pylint to say: `E:309, 8: Assigning to attribute '_unicode' not defined in class slots (assigning-non-slot)`. :-/

opaque_keys/edx/locator.py
ormsbee: @doctoryes: Re-added below.

ormsbee:
@nedbat, @doctoryes: Responded to comments, added tests. Another round please?

ormsbee:
@doctoryes is on vacation this coming week. @jmbowman, @macdiesel: Do one of you folks have time to review this next week?

macdiesel:
@ormsbee code looks good, have you looked at memory usage when this runs on a long soak load test?

ormsbee:
@macdiesel: No, I have not. Do you believe that's a pre-condition to merging?

macdiesel:
Depends, how many keys do you expect there to live in the cache and how long will these items live? Assuming the cache is correctly evicted with the weakref dictionary I don't think it would be a problem, though it's something to keep an eye on post deployment.

macdiesel:
👍

ormsbee:
@macdiesel: Up to 5K items at any given time (usually much less), but shouldn't live much beyond a request. That's verified on devstack by virtue of it getting cache misses, but I haven't sat down and carefully tracked overall memory usage over time. One of the wrinkles is that we are actually creating far fewer objects overall.

Thank you.

opaque_keys/__init__.py
```diff
@@ -168,10 +179,18 @@ def __str__(self):
...
         """
+        # OpaqueKeys are often repeatedly serialized, so we cache this value.
+        if self._unicode is not None:
+            return self._unicode
+
         if self.deprecated:
             # no namespace on deprecated
-            return self._to_deprecated_string()
-        return self.NAMESPACE_SEPARATOR.join([self.CANONICAL_NAMESPACE, self._to_string()])  # pylint: disable=no-member
+            self._unicode = self._to_deprecated_string()  # pylint: disable=assigning-non-slot
```
jmbowman: Are these comments due to a bug in pylint?  (Like [this one](https://github.com/PyCQA/pylint/issues/731), possibly fixable by upgrading python-astroid.)  If so, should we flag these somehow for re-evaluation after we upgrade it?

ormsbee: I believe so. Is there a preferred way to flag it?

jmbowman: I don't know of a particular convention (and I just saw the earlier discussion below about these comments).  Maybe just a comment somewhere in the file like "# TODO Revisit assigning-non-slot comments on pylint upgrade"?  We could then later search the code for "pylint upgrade", or stumble across it when looking for TODO comments.

opaque_keys/edx/locator.py
```diff
@@ -151,6 +153,10 @@ class CourseLocator(BlockLocatorBase, CourseKey):   # pylint: disable=abstract-m
     __slots__ = KEY_FIELDS
     CHECKED_INIT = False
 
+    # A mapping of parsed key fields to CourseLocator objects, for caching.
+    # Every subclass has its own.
+    __parsed_fields_to_keys = WeakValueDictionary()
```
jmbowman: How does every subclass get its own?  I only see the one definition, and thought the double underscore only does name mangling (all subclasses get the same mangled name and object reference, based on the class it was defined in).

ormsbee: Ugh, you're right. That was me misremembering the feature. That makes me very sad. I'll try a couple of alternatives to see how they perform. Thank you. :)

setup.py
```diff
@@ -2,7 +2,7 @@
 
 setup(
     name='edx-opaque-keys',
-    version='0.3.1',
+    version='0.3.2',
```
jmbowman: I don't see a changelog for this package; shouldn't it have one?  (I like to add a comment on this line reminding developers to update it when bumping the version.)

ormsbee:
@jmbowman: Comments addressed. Is this good enough to squash and merge now?

opaque_keys/edx/locator.py
```diff
@@ -260,7 +267,45 @@ def _from_string(cls, serialized):
         if parse['version_guid']:
             parse['version_guid'] = cls.as_object_id(parse['version_guid'])
 
-        return cls(**{key: parse.get(key) for key in cls.KEY_FIELDS})
+        parsed_values = tuple(parse.get(key) for key in cls.KEY_FIELDS)
+
+        # Include the class name in the cache key so that we don't have
+        # collisions with sub-classes.
+        cache_key = (cls.__class__, parsed_values)
```
jmbowman: The comment says class name, the code uses the class object.  The latter is probably more robust (could have multiple classes with the same name in different packages), so just update the comment?

jmbowman:
@ormsbee The changes look good other than that one nitpick.

opaque_keys/edx/locator.py
ormsbee: Fixed.

ormsbee:
Awesome -- I'm going to squash and merge.

PR #73, opened 2016-04-01
Pinning requirements

clintonb:
We definitely need a later version of six to support python_2_unicode_compatible. Pinning the other requirements to avoid potential issues with major version bumps breaking this library.

ECOM-4016

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,14 +2,14 @@ from setuptools import setup, find_packages
 
 setup(
     name='edx-opaque-keys',
-    version='0.3.0',
+    version='0.3.1',
     author='edX',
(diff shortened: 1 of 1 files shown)


clintonb:
@cpennington any idea what minimum versions of `pymongo` and `stevedore` we support?
@brianhw this resolves the issues you were seeing with pipeline
@nedbat thoughts? I know we've discussed requirements pinning a lot in the past.

clintonb:
Chatted with @nedbat in person. Merging as-is. If we need to support earlier versions of `stevedore` or `pymongo` in the future we can do so in another PR.

PR #76, opened 2016-08-24
Protect keys/locators against trailing newlines.

doctoryes:
https://openedx.atlassian.net/browse/ECOM-5345

The key url regular expression was flawed in that it allowed trailing newlines. I haven't found a way yet via regex to detect trailing newlines in the complex regex that the code needs, so a separate check for trailing newlines was added. Tests were also added for other important locators/keys.

@cpennington @ormsbee @cpennington @adampalay @feanil @rlucioni PTAL.

### DIFF
--- a/opaque_keys/edx/locator.py
+++ b/opaque_keys/edx/locator.py
@@ -121,7 +121,20 @@ class BlockLocatorBase(Locator):
         Raises:
             InvalidKeyError: if string cannot be parsed -or- string ends with a newline.
         """
-        # URL_RE above cannot detect a single trailing newline, so guard against them.
+        # URL_RE above cannot detect a single trailing newline, so guard against them
+        # by detecting them separately. An example to demonstrate:
+        #
+        # >>> import re
+        # >>> x = re.compile('(.+)$')
+        # >>> print x.match('foo')
+        # <_sre.SRE_Match object at 0x107df4510>
+        # >>> print x.match('foo\n')
+        # <_sre.SRE_Match object at 0x107df4510>
+        # >>> print x.match('foo\n\n')
+        # None
+        #
+        # The final regex-captured group will *not* capture a trailing newline.
+        # But parse the string strictly with no trailing newlines allowed.
         if string.endswith('\n'):
             raise InvalidKeyError(cls, string)
 
--- a/opaque_keys/edx/locator.py
+++ b/opaque_keys/edx/locator.py
@@ -96,7 +96,7 @@ class BlockLocatorBase(Locator):
         ({BRANCH_PREFIX}@(?P<branch>{ALLOWED_ID_CHARS}+){SEP})?
         ({VERSION_PREFIX}@(?P<version_guid>[A-F0-9]+){SEP})?
         ({BLOCK_TYPE_PREFIX}@(?P<block_type>{ALLOWED_ID_CHARS}+){SEP})?
-        ({BLOCK_PREFIX}@(?P<block_id>{BLOCK_ALLOWED_ID_CHARS}+))?\Z
+        ({BLOCK_PREFIX}@(?P<block_id>{BLOCK_ALLOWED_ID_CHARS}+))?
         """.format(
         ALLOWED_ID_CHARS=Locator.ALLOWED_ID_CHARS,
         BLOCK_ALLOWED_ID_CHARS=BLOCK_ALLOWED_ID_CHARS,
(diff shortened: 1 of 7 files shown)


doctoryes:
@ormsbee We decided to add an obvious loud error instead, so that the data can be corrected.

opaque_keys/edx/locator.py
```diff
@@ -121,8 +121,12 @@ def parse_url(cls, string):
         with key 'id' and optional keys 'branch' and 'version_guid'.
 
         Raises:
-            InvalidKeyError: if string cannot be parsed.
+            InvalidKeyError: if string cannot be parsed -or- string ends with a newline.
```
ormsbee: Does it catch trailing spaces?

opaque_keys/edx/locator.py
```diff
@@ -121,8 +121,12 @@ def parse_url(cls, string):
...
         """
+        # URL_RE above cannot detect a single trailing newline, so guard against them.
```
ormsbee: Does it catch trailing spaces?

doctoryes: The regular expression itself *does* catch trailing spaces. The problem was with the `$` regex operator which, according to the Python `re` docs, does the following: `Matches the end of the string or just before the newline at the end of the string`. The regex was doing the former and parsing the trailing newline just fine.

doctoryes: An example:
```
>>> import re
>>> x = re.compile('.+$')
>>> print x.match('foo')
<_sre.SRE_Match object at 0x107df4510>
>>> print x.match('foo\n')
<_sre.SRE_Match object at 0x107df4510>
>>> print x.match('foo\n\n')
None
```

adampalay:
amazing 👍

rlucioni:
@doctoryes don't forget to bump the package version in setup.py before merging this.

doctoryes:
Thanks, @rlucioni ! I'll do so.

rlucioni:
👍

opaque_keys/edx/locator.py
feanil: Can we put this example block in as a comment as well?

feanil:
One nit otherwise :+1:

opaque_keys/edx/locator.py
doctoryes: Done.

doctoryes:
@ormsbee Are you OK with this PR being merged, taking into account the version bump *and* your eventual re-addition of opaque keys caching?

opaque_keys/edx/locator.py
```diff
@@ -119,8 +119,25 @@ def parse_url(cls, string):
...
-            InvalidKeyError: if string cannot be parsed.
-        """
+            InvalidKeyError: if string cannot be parsed -or- string ends with a newline.
+        """
+        # URL_RE above cannot detect a single trailing newline, so guard against them
+        # by detecting them separately. An example to demonstrate:
+        #
+        # >>> import re
+        # >>> x = re.compile('(.+)$')
+        # >>> print x.match('foo')
```
jibsheet: I mentioned elsewhere that PCRE has \z for matching end of string, not including newline.  Python seems to call that \Z.  Also, . doesn't match \n unless you enable /s (singleline)..  So can you do this as:

```
>>> import re
>>> x = re.compile('(.+)\Z')
>>> print x.match('foo')
<_sre.SRE_Match object at 0x10deeebe8>
>>> print x.match('foo\n')
None
```

doctoryes:
@jibsheet Thanks! Can you please take another look at this version that uses '\Z'?

@cpennington FYI

opaque_keys/edx/locator.py
```diff
@@ -96,15 +96,15 @@ class BlockLocatorBase(Locator):
...
         BLOCK_ALLOWED_ID_CHARS=BLOCK_ALLOWED_ID_CHARS,
         BRANCH_PREFIX=BRANCH_PREFIX,
         VERSION_PREFIX=Locator.VERSION_PREFIX,
         BLOCK_TYPE_PREFIX=Locator.BLOCK_TYPE_PREFIX,
         BLOCK_PREFIX=BLOCK_PREFIX,
-        SEP=r'(\+(?=.)|$)',  # Separator: requires a non-trailing '+' or end of string
+        SEP=r'(\+(?=.)|\Z)',  # Separator: requires a non-trailing '+' or end of string
     )
 
     URL_RE = re.compile('^' + URL_RE_SOURCE + '$', re.IGNORECASE | re.VERBOSE | re.UNICODE)
```
cpennington: Replace `$` with `\Z` here, too?

jibsheet: I wonder if we want to pull the \Z out of 99 and replace the $ on 110 with it, rather than having \Z inside and outside (or \Z inside and $ outside).

doctoryes: Fixed - re-review?

jibsheet:
👍

ormsbee:
👍

PR #77, opened 2016-08-25
Revert "Optimize equality, hashing, serialization."

ormsbee:
This reverts commit a71958ea9b9b4ce1b33dede230bde000caa336e8.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,0 +1,14 @@
+# 0.3.3
+
+* Revert of caching optmizations introduced in 0.3.2, due to a bug that can
+  occur where course keys can be parsed with trailing newlines, and those parsed
+  values can be serialized into the database.
+
+# 0.3.2
+
+* Simple optimizations to reduce the number of OpaqueKey objects
+  created, and to speed up hashing and equality checks.
+
+-----
+
+-No changelog was maintained before 0.3.2.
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -2,7 +2,7 @@
 
 setup(
     name='edx-opaque-keys',
-    version='0.3.2',  # Please update CHANGELOG when you bump versions
+    version='0.3.1',
```
doctoryes: The version needs a bump upward...

CHANGELOG.rst
```diff
@@ -1,8 +0,0 @@
-# 0.3.2
-
-* Simple optimizations to reduce the number of OpaqueKey objects
-  created, and to speed up hashing and equality checks.
-
-----
-
-No changelog was maintained before 0.3.2.
```
doctoryes: We should keep the CHANGELOG and add version 0.3.3?

setup.py
ormsbee: If I bump this right now, this very commit will get uploaded to pypi as 0.3.3. Is that what we want?

doctoryes: It actually won't until we tag the repo - that's when the automatic Travis pkg build/upload would occur.

doctoryes: But - perhaps we should just add any additional 0.3.3 changes into this PR to avoid confusion?

ormsbee: Based on discussion, it sounds like this is indeed what we want.

doctoryes:
👍

PR #84, opened 2017-01-27
Re-applying Optimizations

ormsbee:
Minus the horrible issue that caused a hotfix. The value of `._unicode` always has to be derived from the parsed key, instead of trusting input into `from_string`

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,4 +1,10 @@
-# Unreleased
+# 0.4.1
+
+* Re-introduce the caching optimizations of 0.3.2, but this time don't blindly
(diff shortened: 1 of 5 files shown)


ormsbee:
@cpennington: I think I got rid of the bit that bit us last time (taking the string that came into `from_string` and assigning it directly). Need to fix up the Changelog and commit message, but I think the code is where I want it. I'll give it another look through in the morning.

@nasthagiri: Heads up since this might affect Block Transformer related performance.

ormsbee:
@doctoryes, @macdiesel: Review please?

ormsbee:
@cpennington: Do I need another thumbs on this, or is one sufficient to merge on this repo?

PR #85, opened 2017-02-08
Revert "Re-applying Optimizations"

ormsbee:
@cpennington: The tests passed on the PR, but my change seems to have broken the build on master. Are the tests non-deterministic?

Anyhow, I'm reverting until I can figure it out.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,10 +1,4 @@
-# 0.4.1
-
-* Re-introduce the caching optimizations of 0.3.2, but this time don't blindly
-  assign the argument to OpaqueKey.from_string() to be the key's serialized
-  representation. The OpaqueKey always serializes itself out to whatever its
-  canonical form is. This should be redundant given the other precautions added
-  since 0.3.2, but it's an extra layer of security.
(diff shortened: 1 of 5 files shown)


ormsbee:
https://travis-ci.org/edx/opaque-keys/builds/199478565

cpennington:
The tests use hypothesis, so they will generate new test cases every time. If you hit a test case failure that's interesting, it's good to add it to the set of examples that decorate each test (those will all be checked in every test run).

cpennington:
That said, I don't think the second key should have parsed at all: `i4x://-/-/-/-@-/` (which should be equivalent to `i4x://org/course/category/block_id@version/`. The trailing slash shouldn't be allowed.

ormsbee:
Okay, so picking this up again after a long, long hiatus. :-P I reverted immediately when I saw this failure, but it looks like it's something that happens in master as well when this test case is passed in. Investigating why this isn't rejected.

cpennington:
That's the lovely thing about hypothesis... You can't be sure it's actually found all the things it could possibly find...

PR #86, opened 2017-02-17
Removed pymongo dependency

### LINKED ISSUE #83: pymongo dependency
It looks like opaque-keys lists `pymongo` as a dependency in `setup.py` but I don't see it referenced in the code base. Is this dependency still necessary? Just curious, thanks!

clintonb:
Fixes #83

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,7 +17,6 @@ setup(
     install_requires=[
         'six>=1.10.0,<2.0.0',
         'stevedore>=0.14.1,<2.0.0',
-        'pymongo>=2.7.2,<4.0.0'
     ],
     entry_points={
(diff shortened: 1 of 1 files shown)


clintonb:
@mulby @cpennington @nedbat @ormsbee please review

clintonb:
Drat! We need `pymongo` for `bson`. We cannot install `bson` as a separate requirement because that will conflict with the library installed by `pymongo`. https://github.com/py-bson/bson/issues/29 explains this. It seems py-bson 0.4.5 is supposed to resolve this issue; however, I'm not sure how edx-platform, or other projects that need `pymongo` will be affected. I am closing for now, as I expected this to be a quick fix this morning.

PR #87, opened 2017-03-03
WIP: ENT-211 add aggregate course key

zubair-arbi:
@asadiqbal08 @saleem-latif @mattdrayer 
Add a new type `AggregateCourseKey` of `OpaqueKey` for handling course keys without run information.

### DIFF
--- a/opaque_keys/edx/keys.py
+++ b/opaque_keys/edx/keys.py
@@ -77,6 +77,22 @@ class DefinitionKey(OpaqueKey):
         raise NotImplementedError()
 
 
+class AggregateCourseKey(OpaqueKey):
+    """
+    An :class:`opaque_keys.OpaqueKey` identifying a
+    serialized Course Key object.
+    """
+    KEY_TYPE = 'aggregate_course_key'
+    __slots__ = ()
+
+    @abstractproperty
+    def serialize(self):  # pragma: no cover
+        """
+        The serialized course key without run information.
+        """
+        raise NotImplementedError()
+
+
 class CourseObjectMixin(object):
     """
     An abstract :class:`opaque_keys.OpaqueKey` mixin
--- a/opaque_keys/edx/locator.py
+++ b/opaque_keys/edx/locator.py
@@ -16,7 +16,7 @@ from bson.son import SON
 
 from six import string_types, text_type
 from opaque_keys import OpaqueKey, InvalidKeyError
-from opaque_keys.edx.keys import CourseKey, UsageKey, DefinitionKey, AssetKey
+from opaque_keys.edx.keys import CourseKey, UsageKey, DefinitionKey, AssetKey, AggregateCourseKey
 
 log = logging.getLogger(__name__)
 
@@ -1330,3 +1330,69 @@ class AssetLocator(BlockUsageLocator, AssetKey):    # pylint: disable=abstract-m
 
 # Register AssetLocator as the deprecated fallback for AssetKey
 AssetKey.set_deprecated_fallback(AssetLocator)
+
+
+class AggregateCourseLocator(BlockUsageLocator, AggregateCourseKey):    # pylint: disable=abstract-method
+    """
+    An AggregateCourseKey implementation class.
+    """
+    CANONICAL_NAMESPACE = 'aggregate-course'
+    KEY_FIELDS = ('org', 'course')
+    __slots__ = KEY_FIELDS
+
+    URL_RE_SOURCE = r"""
+        ((?P<org>{ALLOWED_ID_CHARS}+)\+(?P<course>{ALLOWED_ID_CHARS}+))?
+    """.format(
+        ALLOWED_ID_CHARS=Locator.ALLOWED_ID_CHARS
+    )
+
+    URL_RE = re.compile('^' + URL_RE_SOURCE)
+
+    def __init__(self, org=None, course=None, **kwargs):
+        """
+        Construct a AggregateCourseLocator.
+
+        Arguments:
+            org (string): Organization identifier for the course
+            course (string): Course name
+
+        """
+        super(AggregateCourseLocator, self).__init__(
+            org=org,
+            course=course,
+            **kwargs
+        )
+
+        if self.org is None or self.course is None:
+            raise InvalidKeyError(self.__class__, 'Both org and course should be set.')
+
+    @classmethod
+    def _from_string(cls, serialized):
+        """
+        Return a AggregateCourseLocator parsing the given serialized string.
+
+        Arguments:
+            serialized: string for matching
+
+        """
+        parse = cls.URL_RE.match(serialized)
+        if not parse:
+            raise InvalidKeyError(cls, serialized)
+
+        parse = parse.groupdict()
+        return cls(**{key: parse.get(key) for key in cls.KEY_FIELDS})
+
+    def _to_string(self):
+        """
+        Return a string representing this location.
+        """
+        return u"{org}+{course}".format(
+            org=self.org,
+            course=self.course
+        )
+
+    def serialize(self):
+        """
+        The serialized course key without run information.
+        """
+        return '{org}+{course}'.format(org=self.org, course=self.course)
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,7 @@ from setuptools import setup, find_packages
 
 setup(
     name='edx-opaque-keys',
-    version='0.4',
+    version='0.5',
     author='edX',
     url='https://github.com/edx/opaque-keys',
     classifiers=[
@@ -30,6 +30,9 @@ setup(
             'library-v1 = opaque_keys.edx.locator:LibraryLocator',
             # don't use slashes in any new code
             'slashes = opaque_keys.edx.locator:CourseLocator',
+        ],
+        'aggregate_course_key': [
+            'aggregate-course = opaque_keys.edx.locator:AggregateCourseLocator',
         ],
         'usage_key': [
             'block-v1 = opaque_keys.edx.locator:BlockUsageLocator',


zubair-arbi:
@cpennington Please let me know if this implementation is what you have suggested.

opaque_keys/edx/locator.py
```diff
@@ -1330,3 +1330,67 @@ def to_deprecated_list_repr(self):
 
 # Register AssetLocator as the deprecated fallback for AssetKey
 AssetKey.set_deprecated_fallback(AssetLocator)
+
+
+class AggregateCourseLocator(AggregateCourseKey):    # pylint: disable=abstract-method
```
cpennington: What formats of key does this need to parse? Can you add some tests to make sure it parses/serializes the way you expect?

Also, should you be able to get an `AggregateCourseKey` from a `CourseKey`?

opaque_keys/edx/locator.py
```diff
@@ -1330,3 +1330,67 @@ def to_deprecated_list_repr(self):
...
+        return u'{org}+{course}'.format(
+            org=self.org,
+            course=self.course
+        )
+
+    def serialize(self):
+        """
+        The serialized course key without run information.
+        """
+        return '{org}+{course}'.format(org=self.org, course=self.course)
```
zubair-arbi: Removed the method `serialize`

opaque_keys/edx/locator.py
zubair-arbi: @cpennington @clintonb  The only format need is `org+course`. I have added/update tests in `opaque_keys/edx/tests/test_locators.py` for the `AggregateCourseLocator`.

I was unable to directly get a `AggregateCourseKey` locator from a `CourseKey` locator mainly due to changed `KEY_TYPE` and [`OpaqueKey` Hierarchy](https://github.com/edx/edx-platform/wiki/Opaque-Keys-(Locators)#opaquekey-hierarchy). To handle the conversion from `CourseKey` to `AggregateCourseKey` I have added a new method `from_course_key` in `AggregateCourseKey` which takes a course key object and returns a locator `AggregateCourseLocator` from it. Please let me know your suggestion on this approach.

opaque_keys/edx/tests/test_locators.py
```diff
@@ -59,3 +61,44 @@ def test_version_tree(self):
...
+        aggregate_course_locator = AggregateCourseLocator.from_course_key(course)
+
+        aggregate_course_key = 'aggregate-course:{}+{}'.format(course.org, course.course)
+        self.assertEqual(aggregate_course_locator, AggregateCourseKey.from_string(aggregate_course_key))
+
+    def test_aggregate_course_locator_serialize(self):
+        aggregate_course_key = 'aggregate-course:org+course'
+        aggregate_course_locator = AggregateCourseKey.from_string(aggregate_course_key)
+        serialized_aggregate_course_key = 'org+course'
+        # Allow access to _to_string
```
clintonb: This comment isn't necessary.

setup.py
```diff
@@ -2,7 +2,7 @@
 
 setup(
     name='edx-opaque-keys',
-    version='0.4',
+    version='0.5',
```
clintonb: 0.5.0

opaque_keys/edx/tests/test_locators.py
```diff
@@ -59,3 +61,44 @@ def test_version_tree(self):
...
         self.assertEqual(VersionTree(valid_locator).children, [])
+
+
+@ddt.ddt
+class AggregateCourseLocatorTests(TestCase):
+    """
+    Tests for :class:`.AggregateCourseLocator`
+    """
+
+    def test_aggregate_course_locator(self):
```
clintonb: Add docstrings explaining what's being tested in each of these methods.

opaque_keys/edx/tests/test_locators.py
```diff
@@ -59,3 +61,44 @@ def test_version_tree(self):
...
+        aggregate_course_key = 'aggregate-course:org+course'
+        aggregate_course_locator = AggregateCourseLocator(org='org', course='course')
+        self.assertEqual(aggregate_course_locator, AggregateCourseKey.from_string(aggregate_course_key))
+
+    @ddt.data(
+        'org/course/run',
+        'course-v1:org+course+run',
+    )
+    def test_aggregate_course_locator_from_course_key(self, course_id):
+        course = CourseKey.from_string(course_id)
```
clintonb: s/course/course_key

opaque_keys/edx/tests/test_locators.py
```diff
@@ -59,3 +61,44 @@ def test_version_tree(self):
...
+        aggregate_course_locator = AggregateCourseLocator.from_course_key(course)
+
+        aggregate_course_key = 'aggregate-course:{}+{}'.format(course.org, course.course)
+        self.assertEqual(aggregate_course_locator, AggregateCourseKey.from_string(aggregate_course_key))
```
clintonb: The use of `AggregateCourseKey.from_string` is confusing and unnecessary.

```python
course_key = CourseKey.from_string(course_id)
expected = AggregateCourseKey(org=course_key.org, course=course_key.course)
actual = AggregateCourseKey.from_course_key(course_key)
assert expected == actual
```

opaque_keys/edx/tests/test_locators.py
```diff
@@ -59,3 +61,44 @@ def test_version_tree(self):
...
+
+    def test_aggregate_course_locator_serialize(self):
```
clintonb: `test_serialize_to_string`

opaque_keys/edx/tests/test_locators.py
```diff
@@ -59,3 +61,44 @@ def test_version_tree(self):
...
+        aggregate_course_key = 'aggregate-course:org+course'
+        aggregate_course_locator = AggregateCourseKey.from_string(aggregate_course_key)
```
clintonb: You should not need to use `AggregateCourseKey.from_string` to instantiate a new instance of `AggregateCourseKey`. Fix this here, and in all other tests where it is used unnecessarily.

opaque_keys/edx/locator.py
```diff
@@ -1330,3 +1330,72 @@ def to_deprecated_list_repr(self):
...
+            course (string): Course name
+
+        """
+        super(AggregateCourseLocator, self).__init__(
+            org=org,
+            course=course,
+            **kwargs
+        )
+
+        if self.org is None or self.course is None:
```
clintonb: You need to account for empty string: `if not (self.org and self.course)`

opaque_keys/edx/locator.py
```diff
@@ -1330,3 +1330,72 @@ def to_deprecated_list_repr(self):
...
+            raise InvalidKeyError(cls, serialized)
+
+        parse = parse.groupdict()
+        return cls(**{key: parse.get(key) for key in cls.KEY_FIELDS})
+
+    def _to_string(self):
+        """
+        Return a string representing this location.
+        """
+        return u'{org}+{course}'.format(
```
clintonb: No need for the `u`.

opaque_keys/edx/keys.py
```diff
@@ -77,6 +77,26 @@ def block_type(self):  # pragma: no cover
         raise NotImplementedError()
 
 
+class AggregateCourseKey(OpaqueKey):
+    """
+    An :class:`opaque_keys.OpaqueKey` identifying a
+    serialized Course Key object.
+    """
+    KEY_TYPE = 'aggregate_course_key'
```
cpennington: No technical issues that I can think of.

mattdrayer: Is this something we should do as part of this PR?  We can certainly make the change but I don't want to introduce two new ideas at one time if we don't have to

opaque_keys/edx/locator.py
```diff
@@ -1330,3 +1330,72 @@ def to_deprecated_list_repr(self):
...
+
+    URL_RE = re.compile(URL_RE_SOURCE)
+
+    def __init__(self, org=None, course=None, **kwargs):
+        """
+        Construct a AggregateCourseLocator.
+
+        Arguments:
+            org (string): Organization identifier for the course
+            course (string): Course name
```
zubair-arbi: My understanding is we call the middle part of key as course or course name `organization+course_name+course_run` and all together are considered as a unique identifier for a course. I can change it if its confusing.

opaque_keys/edx/keys.py
clintonb: I would call this course-v2 now. Add a ticket to your backlog to add course-run-v1.

zubair-arbi: @cpennington @clintonb  I have to change the key type to `course_key_v2` or just `course_v2` because the `CANONICAL_NAMESPACE` of `AggregateCourseLocator` would be `course-v2`? Also should I change the name of aggregate course key and locator names too?

opaque_keys/edx/tests/test_locators.py
zubair-arbi: added

opaque_keys/edx/locator.py
zubair-arbi: I have changed it to `Course number` as per `Studio`.

opaque_keys/edx/tests/test_locators.py
zubair-arbi: @clintonb @cpennington This seems to be the [documented way](https://github.com/edx/edx-platform/wiki/Opaque-Keys-(Locators)#parse-course-ids) if you try to initialize the `AggregateCourseKey` directly you get error:
```python
>>> AggregateCourseKey(org=course_key.org, course=course_key.course)
TypeError: Can't instantiate abstract class AggregateCourseKey with abstract
methods _to_string, from_course_key
```

opaque_keys/edx/keys.py
zubair-arbi: I have change the `CANONICAL_NAMESPACE` of `AggregateCourseLocator` to `course-v2`. Now the aggregate course key format would be `course-v2:org+course`
```python
AggregateCourseKey.from_string('course-v2:edX+DemoX')
```

zubair-arbi:
@cpennington @clintonb please give another review.

opaque_keys/edx/tests/test_locators.py
```diff
@@ -59,3 +61,59 @@ def test_version_tree(self):
...
+
+@ddt.ddt
+class AggregateCourseLocatorTests(TestCase):
+    """
+    Tests for :class:`.AggregateCourseLocator`
+    """
+
+    def test_aggregate_course_locator(self):
+        """
+        Verify that the method "from_string" of class "AggregateCourseKey"
```
clintonb: Use ` instead of " for Python code (classes, methods).

opaque_keys/edx/keys.py
clintonb: I thought we agreed to get rid of the word aggregate. I have no clue what it means in this context.

zubair-arbi:
@clintonb @cpennington I have updated the naming. Please review.

opaque_keys/edx/keys.py
```diff
@@ -77,6 +77,26 @@ def block_type(self):  # pragma: no cover
...
+class CourseKeyV2(OpaqueKey):
+    """
+    An :class:`opaque_keys.OpaqueKey` identifying a
+    serialized Course Key object.
+    """
+    KEY_TYPE = 'course_key_v2'
+    __slots__ = ()
+
+    @abstractproperty
+    def from_course_key(self, course_key):  # pragma: no cover
```
clintonb: Call this `from_course_run_key`. Eventually, course-key-v1 will be replaced with course-run-key-v1.

opaque_keys/edx/tests/test_locators.py
```diff
@@ -59,3 +61,59 @@ def test_version_tree(self):
...
+
+@ddt.ddt
+class CourseLocatorV2Tests(TestCase):
+    """
+    Tests for :class:`.CourseLocatorV2`
+    """
+
+    def test_course_locator_v2(self):
+        """
+        Verify that the method "from_string" of class "CourseKeyV2"
```
clintonb: Use ` instead of " to format Python entities. This should be fixed in all of your docstrings in this file.

opaque_keys/edx/tests/test_locators.py
```diff
@@ -59,3 +61,59 @@ def test_version_tree(self):
...
+        returns an object of "CourseLocatorV2" for a valid course key.
+        """
+        course_key_v2 = 'course-v2:org+course'
+        course_locator_v2 = CourseKeyV2.from_string(course_key_v2)
+        expected_course_locator = CourseLocatorV2(org='org', course='course')
+        self.assertEqual(expected_course_locator, course_locator_v2)
+
+    @ddt.data(
+        'org/course/run',
```
clintonb: There is no need for `ddt` in this instance.  Whether we use the old or new format for `CourseKey` doesn't matter. The method simply expects an instance of `CourseKey`.

opaque_keys/edx/tests/test_locators.py
```diff
@@ -59,3 +61,59 @@ def test_version_tree(self):
...
+
+    def test_serialize_to_string(self):
+        """
+        Verify that the method "_to_string" of class "CourseLocatorV2"
+        serializes a course key v2 to a string with expected format.
+        """
+        course_key = CourseKey.from_string('course-v1:org+course+run')
+        course_locator_v2 = CourseLocatorV2(org=course_key.org, course=course_key.course)
+        expected_serialized_key = '{org}+{course}'.format(org=course_key.org, course=course_key.course)
+        # pylint: disable=protected-access
```
clintonb: No need to call the protected member. Use `str(course_locator_v2)`.

opaque_keys/edx/tests/test_locators.py
```diff
@@ -59,3 +61,59 @@ def test_version_tree(self):
...
+        expected_course_key = CourseLocatorV2(org=course_key.org, course=course_key.course)
+        actual_course_key = CourseLocatorV2.from_course_key(course_key)
+        self.assertEqual(expected_course_key, actual_course_key)
+
+    def test_serialize_to_string(self):
+        """
+        Verify that the method "_to_string" of class "CourseLocatorV2"
+        serializes a course key v2 to a string with expected format.
+        """
+        course_key = CourseKey.from_string('course-v1:org+course+run')
```
clintonb: This is unnecessary.

opaque_keys/edx/tests/test_locators.py
```diff
@@ -59,3 +61,59 @@ def test_version_tree(self):
...
         self.assertEqual(VersionTree(valid_locator).children, [])
+
+
+@ddt.ddt
+class CourseLocatorV2Tests(TestCase):
+    """
+    Tests for :class:`.CourseLocatorV2`
+    """
+
+    def test_course_locator_v2(self):
```
clintonb: `test_from_string`

opaque_keys/edx/tests/test_locators.py
```diff
@@ -59,3 +61,59 @@ def test_version_tree(self):
...
+        course_key_v2 = 'course-v2:org+course'
+        course_locator_v2 = CourseKeyV2.from_string(course_key_v2)
+        expected_course_locator = CourseLocatorV2(org='org', course='course')
+        self.assertEqual(expected_course_locator, course_locator_v2)
+
+    @ddt.data(
+        'org/course/run',
+        'course-v1:org+course+run',
+    )
+    def test_course_locator_v2_from_course_key(self, course_id):
```
clintonb: You don't need to prefix your test methods with `course_locator_v2`. The fact that `CourseLocatorV2` is given by the name of the test class, `CourseLocatorV2Tests`. This should simply be `test_from_course_key`.

opaque_keys/edx/tests/test_locators.py
```diff
@@ -59,3 +61,59 @@ def test_version_tree(self):
...
+        # pylint: disable=protected-access
+        self.assertEqual(expected_serialized_key, course_locator_v2._to_string())
+
+    @ddt.data(
+        'org/course/run',
+        'org+course+run',
+        'org+course+run+foo',
+        'course-v2:org+course+run',
+    )
+    def test_invalid_format_course_key(self, course_key):
```
clintonb: `test_from_string_with_invalid_input`

zubair-arbi:
@clintonb Please have another look.
@cpennington I have added inter-conversion methods for course key and course key v2. I have updated the sample usage in the description of this PR. Since the `LibraryLocator` is course agnostic so the method `make_course_key_v2` will just raise error `NotImplementedError`.

opaque_keys/edx/locator.py
zubair-arbi: a) Sorry I forgot to mention that the complete format of course key has version, e.g. `course-v2:edX+DemoX`
b) I have added the method `make_course_key_v2` in `CourseKey` which will return `CourseKeyV2` locator. Also added the method `make_course_run_key` in `CourseKeyV2` which will return a `CourseKey` locator `CourseLocator` from the provided course run.
@cpennington thanks for the suggestions. Please have a look on the changes and let me know if I have missed anything.

opaque_keys/edx/keys.py
```diff
@@ -61,6 +61,13 @@ def make_asset_key(self, asset_type, path):  # pragma: no cover
         """
         raise NotImplementedError()
 
+    @abstractmethod
+    def make_course_key_v2(self):  # pragma: no cover
```
clintonb: Is this really necessary given that you already have `CourseKeyV2.from_course_run_key()`?

opaque_keys/edx/keys.py
```diff
@@ -77,6 +84,37 @@ def block_type(self):  # pragma: no cover
...
+        Get course key v2 from the course run key.
+
+        Arguments:
+            course_key (:class:`CourseKey`): The course identifier.
+
+        """
+        raise NotImplementedError()
+
+    @abstractproperty
+    def make_course_run_key(self, course_run):  # pragma: no cover
```
clintonb: Why is this needed? We already have existing methods of creating course run keys.

cpennington: These were my suggestion. The existing keys typically are related by giving the more general key a way to create more specific keys by hydrating additional information.

cpennington: That said, I agree that it doesn't make much sense to have both methods. We should pick one, and stick to it.

opaque_keys/edx/keys.py
zubair-arbi: @cpennington @clintonb just to confirm, I have to remove this method only or do I need to remove the `make_course_run_key` in class `CourseKeyV2` too?

cpennington: @zubair-arbi: You should have some a method to go from `CourseKeyV2` to `CourseKey` and back again, but only one of each. I would suggest using regular methods, rather than class methods, in general. So, probably `CourseKeyV2.make_course_run_key`, and `CourseKey.make_course_key_v2`.

cpennington:
I'm concerned about the names `CourseKeyV2` and `CourseKey`. They are fundamentally different things, but that distinction isn't really captured in the names. I think it would be awesome to deprecate `CourseKey` and start encouraging the use of `CourseRunKey` and `AggregateCourseKey` (which are clearly different). "Course" is just too ambiguous right now, until a vast majority of our software switches to using CourseRun for course runs.

PR #88, opened 2017-07-20
Optimization, and some questions about serialization roundtrips

ormsbee:
I reverted the optimizations again in #85 because hypothesis failed a master build on `i4x://-/-/-/-@-/`, which should not have parsed but does. Picking it up again months later, I see that this particular string fails on master as well.

Instead of trying to tweak the regular expression to flag this particular case as invalid, I did a big hammer approach of always doing a serialization roundtrip when calling `OpaqueKey.from_string()` -- it will now raise an `InvalidKeyError` whenever a key does not reserialize into the same representation it was parsed from. This caused a few other tests to fail, and I wanted to discuss what our desired behavior is here (the "Opaque Key Laws" that @cpennington has talked about codifying in the past).

I haven't measured the performance impact of doing this, and it may not be practical to keep in. But I thought it would raise some interesting examples to talk about.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,4 +1,10 @@
-# Unreleased
+# 0.4.1
+
+* Re-introduce the caching optimizations of 0.3.2, but this time don't blindly
+  assign the argument to OpaqueKey.from_string() to be the key's serialized
+  representation. The OpaqueKey always serializes itself out to whatever its
+  canonical form is. This should be redundant given the other precautions added
+  since 0.3.2, but it's an extra layer of security.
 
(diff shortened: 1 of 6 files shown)


ormsbee:
FYI: @nasthagiri

ormsbee:
Okay, this might actually be due to how I'm doing the parsed value caching in CourseLocator. >_< I'll look at it tomorrow.

ormsbee:
Okay, verified this morning that this is not due to the caching code. I'm going to open a separate PR with just the relevant commits to talk about this...

PR #89, opened 2017-07-20
Serialization Roundtrip Issues

ormsbee:
I reverted the optimizations again in #85 because hypothesis failed a master build on `i4x://-/-/-/-@-/`, which should not have parsed but does. Picking it up again months later, I see that this particular string fails on master as well.

Instead of trying to tweak the regular expression to flag this particular case as invalid, I did a big hammer approach of always doing a serialization roundtrip when calling OpaqueKey.from_string() -- it will now raise an InvalidKeyError whenever a key does not reserialize into the same representation it was parsed from. This fixed the original error, but caused a few other tests to fail, and I wanted to discuss what our desired behavior is here (the "Opaque Key Laws" that @cpennington has talked about codifying in the past).

I haven't measured the performance impact of doing this, and it may not be practical to keep in. But I thought it would raise some interesting examples to talk about.

### DIFF
--- a/opaque_keys/__init__.py
+++ b/opaque_keys/__init__.py
@@ -207,7 +207,7 @@ class OpaqueKey(with_metaclass(OpaqueKeyMetaclass)):
         # such a value to pollute the cache.
         reserialized = text_type(key)
         if serialized != reserialized:
-            raise InvalidKeyError(cls, "{} -> {}".format(serialized, reserialized))
+            raise InvalidKeyError(cls, u"{} -> {}".format(serialized, reserialized))
 
         cls._cache_pool[serialized] = key
         return key
--- a/opaque_keys/__init__.py
+++ b/opaque_keys/__init__.py
@@ -189,13 +189,28 @@ class OpaqueKey(with_metaclass(OpaqueKeyMetaclass)):
         # pylint: disable=protected-access
         # load drivers before checking for attr
         cls._drivers()
-        try:
-            namespace, rest = cls._separate_namespace(serialized)
-            return cls.get_namespace_plugin(namespace)._from_string(rest)
-        except InvalidKeyError:
-            if hasattr(cls, 'deprecated_fallback'):
-                return cls.deprecated_fallback._from_deprecated_string(serialized)
-            raise InvalidKeyError(cls, serialized)
+
+        def _parse_with_fallback():
+            """Try to parse as new style key, then fall back to old style."""
+            try:
+                namespace, rest = cls._separate_namespace(serialized)
+                return cls.get_namespace_plugin(namespace)._from_string(rest)
+            except InvalidKeyError:
+                if hasattr(cls, 'deprecated_fallback'):
+                    return cls.deprecated_fallback._from_deprecated_string(serialized)
+                raise InvalidKeyError(cls, serialized)
+
+        key = _parse_with_fallback()
+
+        # We've been bitten by keys that do not re-serialize to the same value
+        # they were parsed from. This is a guard to make sure we never allow
+        # such a value to pollute the cache.
+        reserialized = text_type(key)
+        if serialized != reserialized:
+            raise InvalidKeyError(cls, "{} -> {}".format(serialized, reserialized))
+
+        cls._cache_pool[serialized] = key
+        return key
 
     @classmethod
     def _separate_namespace(cls, serialized):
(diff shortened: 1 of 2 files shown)


ormsbee:
These are the failures:

```
course-v1:version@519665f6223ebd6980884f2b+block@hw3
course-v1:version@519665f6223ebd6980884f2b
```

```
slashes:DemoUniversity+DM01+2014
course-v1:DemoUniversity+DM01+2014
```

```
library-v1:version@519665f6223ebd6980884f2b+block@hw3
library-v1:version@519665f6223ebd6980884f2b
```

These keys cannot be made directly (e.g. `CourseKey.from_string('course-v1:version@519665f6223ebd6980884f2b+block@hw3')` already raises `InvalidKeyError`) so I suspect these keys are made in the process of parsing full `UsageKey` instances that include courses/libraries in them.

ormsbee:
FYI: @nasthagiri

ormsbee:
Okay, more confusion... when I do:

```python
CourseKey.from_string('course-v1:version@519665f6223ebd6980884f2b+block@hw3')
```

It blows up in the parsing. But when a test_course_locators.py:test_course_constructor_url does the same thing, it explodes only with the reserialization check.

ormsbee:
Bleh, it was a combination of invoking a bpython on my path that had similar but not the same libs installed and some pyc caching issues. But using master, I see the roundtrip behavior above.

```
>>> from opaque_keys.edx.keys import CourseKey
>>> ck = CourseKey.from_string("course-v1:version@519665f6223ebd6980884f2b+block@hw3")
>>> str(ck)
'course-v1:version@519665f6223ebd6980884f2b'
>>> sk = CourseKey.from_string("slashes:DemoUniversity+DM01+2014")
>>> str(sk)
'course-v1:DemoUniversity+DM01+2014'
```

So I think that prompts a couple of questions:

1. Have we moved past the point where we no longer need to support "slashes:"? Honestly, I didn't even remember that one existed.
2. Should we disallow the silent truncation of `+block@hw3` in the example above, and flag it as an `InvalidKey` from the start?

ormsbee:
To expand on that a little further, we explicitly wrote a test to make sure `course-v1:version@519665f6223ebd6980884f2b+block@hw3` would parse, but I don't know what the reasoning was.

cpennington:
Looking at https://github.com/edx/opaque-keys/blob/master/opaque_keys/edx/locator.py#L146-L162, it doesn't look like CourseLocator should ever be able to take a `block` argument. And I'm not sure what it would mean if it did.

Looking at the change that introduced that test, I can't see anything in there that convinces me that we shouldn't just make that case an error.

ormsbee:
@cpennington, @dianakhuang: Do either of you remember what the story was behind `slash:` keys?

cpennington:
They were an aborted attempt at backwards compatibility. I'm not 100% sure, but I think they never actually made it to prod.

ormsbee:
Okay, so my plan is:

1. Remove support for `slash:` as a key type -- just delete that mapping entirely.
2. Enforce that keys must successfully roundtrip in their serialization to be considered valid. Change the test code to reflect this where necessary.
3. Re-introduce my performance optimizations.

Items 2 and 3 on this list may have to be coupled for performance reasons, as the roundtrip check will cause some overhead. On the bright side, once we enforce that constraint, we may be able to make hashing more performant. I need to measure it to be sure.

Item 1 can be done as a standalone thing. Given that we'd explicitly be breaking compatibility there, would there be any objection to doing the removal and declaring opaque-keys to be v 1.0?

ormsbee:
@edx/platform-team: As the owners of opaque-keys, does the above sound good to you? ^

ormsbee:
As a sidenote, making this guarantee could potentially allow for faster hashing and equality checking, since we could lean more heavily on the serialized string for those (fewer objects to compare and construct), though that's a secondary concern at this point.

ormsbee:
FYI @pwnage101

doctoryes:
I'm definitely in favor of removing `slash:`.

Allow me to re-state what you're proposing to see if I understand it:
- When an OpaqueKey is converted from string A to object to string B, the strings A & B will be equivalent.
- When an OpaqueKey is converted from object X to string to object Y, the objects X & Y will be equivalent.
- Equivalent objects will always serialize to the same string. Inequivalent objects will *never* serialize to the same string.

The above was my assumption on how opaque keys worked to begin with - and probably how most people thought they worked. So I'm in favor of enforcing these principles in code and communicating them out to all.

ormsbee:
@doctoryes: Yup, that's the intention!

PR #90, opened 2018-01-08
Extract OpaqueKeyFields from edx-platform.

iloveagent57:
https://openedx.atlassian.net/browse/EDUCATOR-2011

The Educator team would like to move the code in lms/djangoapps/completion to its own repository (and then pip install it into edx-platform).  The completion models depend on OpaqueKeyFields for tracking course ids and block ids.  This pull request copies the code in `openedx.core.djangoapps.xmodule_django.models` and `openedx.core.djangoapps.util.model_utils`.  This allows us to de-couple the dependency on edx-platform for the completion model definitions.

### DIFF
--- a/opaque_keys/edx/django/models.py
+++ b/opaque_keys/edx/django/models.py
@@ -1,0 +1,209 @@
+"""
+Useful django models for implementing XBlock infrastructure in django.
+If Django is unavailable, none of the classes below will work as intended.
+"""
+import logging
+import warnings
+
+try:
+    from django.core.exceptions import ValidationError
+    from django.db.models import CharField
+    from django.db.models.lookups import IsNull
+except ImportError:  # pragma: no cover
+    # Django is unavailable, none of the classes below will work,
+    # but we don't want the class definition to fail when interpreted.
+    CharField = object
+    IsNull = object
+
+import six
+
+from opaque_keys.edx.keys import BlockTypeKey, CourseKey, UsageKey
+
+
+log = logging.getLogger(__name__)
+
+
+class _Creator(object):
+    """
+    DO NOT REUSE THIS CLASS. Provided for backwards compatibility only!
+
+    A placeholder class that provides a way to set the attribute on the model.
+    """
+    def __init__(self, field):
+        self.field = field
+
+    def __get__(self, obj, type=None):  # pylint: disable=redefined-builtin
+        if obj is None:
+            return self  # pragma: no cover
+        return obj.__dict__[self.field.name]
+
+    def __set__(self, obj, value):
+        obj.__dict__[self.field.name] = self.field.to_python(value)
+
+
+# pylint: disable=missing-docstring,unused-argument
+class CreatorMixin(object):
+    """
+    Mixin class to provide SubfieldBase functionality to django fields.
+    See: https://docs.djangoproject.com/en/1.11/releases/1.8/#subfieldbase
+    """
+    def contribute_to_class(self, cls, name, *args, **kwargs):
+        super(CreatorMixin, self).contribute_to_class(cls, name, *args, **kwargs)
+        setattr(cls, name, _Creator(self))
+
+    def from_db_value(self, value, expression, connection, context):
+        return self.to_python(value)
+
+
+def _strip_object(key):
+    """
+    Strips branch and version info if the given key supports those attributes.
+    """
+    if hasattr(key, 'version_agnostic') and hasattr(key, 'for_branch'):
+        return key.for_branch(None).version_agnostic()
+    else:
+        return key
+
+
+def _strip_value(value, lookup='exact'):
+    """
+    Helper function to remove the branch and version information from the given value,
+    which could be a single object or a list.
+    """
+    if lookup == 'in':
+        stripped_value = [_strip_object(el) for el in value]
+    else:
+        stripped_value = _strip_object(value)
+    return stripped_value
+
+
+# pylint: disable=logging-format-interpolation
+class OpaqueKeyField(CreatorMixin, CharField):
+    """
+    A django field for storing OpaqueKeys.
+
+    The baseclass will return the value from the database as a string, rather than an instance
+    of an OpaqueKey, leaving the application to determine which key subtype to parse the string
+    as.
+
+    Subclasses must specify a KEY_CLASS attribute, in which case the field will use :meth:`from_string`
+    to parse the key string, and will return an instance of KEY_CLASS.
+    """
+    description = "An OpaqueKey object, saved to the DB in the form of a string."
+
+    Empty = object()
+    KEY_CLASS = None
+
+    def __init__(self, *args, **kwargs):
+        if self.KEY_CLASS is None:
+            raise ValueError('Must specify KEY_CLASS in OpaqueKeyField subclasses')
+
+        super(OpaqueKeyField, self).__init__(*args, **kwargs)
+
+    def to_python(self, value):
+        if value is self.Empty or value is None:
+            return None
+
+        error_message = "%s is not an instance of six.string_types or %s" % (value, self.KEY_CLASS)
+        assert isinstance(value, six.string_types + (self.KEY_CLASS,)), error_message
+        if value == '':
+            # handle empty string for models being created w/o fields populated
+            return None
+
+        if isinstance(value, six.string_types):
+            if value.endswith('\n'):
+                # An opaque key with a trailing newline has leaked into the DB.
+                # Log and strip the value.
+                log.warning(u'{}:{}:{}:to_python: Invalid key: {}. Removing trailing newline.'.format(
+                    self.model._meta.db_table,  # pylint: disable=protected-access
+                    self.name,
+                    self.KEY_CLASS.__name__,
+                    repr(value)
+                ))
+                value = value.rstrip()
+            return self.KEY_CLASS.from_string(value)
+        else:
+            return value
+
+    def get_prep_value(self, value):
+        if value is self.Empty or value is None:
+            return ''  # CharFields should use '' as their empty value, rather than None
+
+        assert isinstance(value, self.KEY_CLASS), "%s is not an instance of %s" % (value, self.KEY_CLASS)
+        serialized_key = six.text_type(_strip_value(value))
+        if serialized_key.endswith('\n'):
+            # An opaque key object serialized to a string with a trailing newline.
+            # Log the value - but do not modify it.
+            log.warning(u'{}:{}:{}:get_prep_value: Invalid key: {}.'.format(
+                self.model._meta.db_table,  # pylint: disable=protected-access
+                self.name,
+                self.KEY_CLASS.__name__,
+                repr(serialized_key)
+            ))
+        return serialized_key
+
+    def validate(self, value, model_instance):
+        """Validate Empty values, otherwise defer to the parent"""
+        # raise validation error if the use of this field says it can't be blank but it is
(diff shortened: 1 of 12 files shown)


opaque_keys/models.py
```diff
@@ -0,0 +1,226 @@
...
+    """
+    def contribute_to_class(self, cls, name, *args, **kwargs):
+        super(CreatorMixin, self).contribute_to_class(cls, name, *args, **kwargs)
+        setattr(cls, name, Creator(self))
+
+    def from_db_value(self, value, expression, connection, context):
+        return self.to_python(value)
+
+
+class NoneToEmptyManager(models.Manager):
```
macdiesel: I don't see this used anywhere else in your PR.

opaque_keys/models.py
```diff
@@ -0,0 +1,226 @@
...
+
+    def __set__(self, obj, value):
+        obj.__dict__[self.field.name] = self.field.to_python(value)
+
+
+# pylint: disable=missing-docstring,unused-argument
+class CreatorMixin(object):
+    """
+    Mixin class to provide SubfieldBase functionality to django fields.
+    See: https://docs.djangoproject.com/en/1.11/releases/1.8/#subfieldbase
```
macdiesel: Please add this to this comment:

DO NOT REUSE THIS CLASS. Provided for backwards compatibility only!

brianhw:
I am concerned about the introduction of a dependency on Django for this.  We use this in the analytics pipeline, and really don't want to have this pulling in all of Django with it, only to be able to parse an opaque key we encounter in a data file.

iloveagent57:
@macdiesel Mind giving this another run-through when you're free?

iloveagent57:
Also, here's a ticket to do the work to extract code from edx-platform: https://openedx.atlassian.net/browse/EDUCATOR-2086

iloveagent57:
@edx/educator-neem

iloveagent57:
@nasthagiri @jmbowman @cpennington @ormsbee would one (or more) of you mind giving this a review?  Thanks!

ormsbee:
@iloveagent57: I can review this in the afternoon today.

cpennington:
You could have a separate tox context that doesn't install django.

tox.ini
```diff
@@ -1,9 +1,14 @@
 [tox]
-envlist = py27,py35
+envlist = py{27,35},django18
```
macdiesel: We should also test this with Django 1.11

iloveagent57:
@schenedx Just running `tox` should run unit tests both with and without Django installed in a virtualenv.  If you want to do that locally: `tox -e py27` (or py35 if you're using python 3).

tox.ini
iloveagent57: Good point.

iloveagent57:
@ormsbee (or anyone else on this thread) any chance you'll be able to look at this today?

settings/base.py
```diff
@@ -0,0 +1,27 @@
...
+ROOT_URLCONF = 'urls'
+
+INSTALLED_APPS = (
+    'django.contrib.auth',
+    'django.contrib.contenttypes',
+)
+
+BASE_DIR = os.path.abspath(os.path.dirname(os.path.dirname(__file__)))
+
+LOCALE_PATHS = [os.path.join(BASE_DIR, "openassessment", "locale")]
```
ormsbee: Why is this line here?

settings/base.py
```diff
@@ -0,0 +1,27 @@
...
+    }
+}
+
+SECRET_KEY = 'the-secret-key'
+
+ROOT_URLCONF = 'urls'
+
+INSTALLED_APPS = (
+    'django.contrib.auth',
+    'django.contrib.contenttypes',
```
ormsbee: Is even this necessary?

opaque_keys/edx/django/models.py
```diff
@@ -0,0 +1,209 @@
...
+    KEY_CLASS = CourseKey
+
+
+class UsageKeyField(OpaqueKeyField):
+    """
+    A django Field that stores a UsageKey object as a string.
+    """
+    description = "A Location object, saved to the DB in the form of a string"
+    KEY_CLASS = UsageKey
+
```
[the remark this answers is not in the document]
iloveagent57: Yeah, that's the idea - I didn't want to start conditionally importing things from edx-platform.

settings/base.py
iloveagent57: No, this can probably be an empty tuple.

settings/base.py
iloveagent57: copypasta!

iloveagent57:
I successfully ran the tubular tests locally against a locally installed version 0.4.2 of opaque-keys (i.e. this PR).  @ormsbee mind giving this the ole seal of approval?  Or would you prefer that we run that structures script against stage first?

nasthagiri:
I realize this PR is now merged and closed, but I'm bringing up an earlier thread since I'd like to close it given it may be a design pattern we'll continue to use.

> So the caveat on this review is that I disagree with the approach of folding Django dependencies into opaque-keys. Honestly, I don't even like that we put pymongo in here. Not because of the weight of the dependency, but because I don't like try-except-ImportError magic, and I think it's clearer and more explicit to keep the Django integration logic separate from the keys, since there are projects that don't care about Django at all (e.g. ccx-keys, some pipeline code, splitmongo cleanup code in tubular, etc.). In general, I favor simple and explicit even if it's slightly less convenient packaging.

I agree with keeping unrelated logic separate (per [Interface Segregation](https://en.wikipedia.org/wiki/Interface_segregation_principle)).  In this case, I think the Django storage of opaque keys is a functionality that is closely related to opaque-keys and would need to change if/when we add new `KEY_CLASS`es, etc.  So, keeping it within this repo makes sense (per [Single Responsibility](https://en.wikipedia.org/wiki/Single_responsibility_principle)).

I'd like to go forward thinking about the placement of code based on its responsibilities and domain ownership rather than its technology dependencies and constraints.  I'm concerned that our current technology-based division of relatively hard to navigate (`lib`, `djangoapps`, `djangolib`, etc) folders within the monolith will transfer to across repos - which can grow into an even larger scale problem. - unless we begin to think about domains/concepts/ownership/responsibilities.

Yes, there are technical limitations with how django/python dependencies work.  However, I believe the `extras_require` solution is a good approach and seems to be a standard way of accomplishing this.

Also, I think we can/should now remove the `try-except ImportError` in the code.  This shouldn't be needed if non-Django clients of Opaque-keys just don't use the django module in this repo, right?  @iloveagent57 ?

So what do you think about this mindset @ormsbee ? 

CCing others who may have opinions on this: @cpennington @nedbat @robrap

ormsbee:
Okay, let me preface this entire comment by saying that I didn't feel strongly enough about this issue to contest it, which is why I gave a thumbs up on this PR. There are tradeoffs either way, and I have a mild preference for separation.

I'm not familiar with the details of how `extras_require` is used in Python. I mean, I know what it is and why people do it at a high level. But I don't know what best practices are around developing, testing, and making sure it works properly in places where those dependencies do and don't exist. Whenever I see "import this, catch ImportError, and substitute things with object()", I think that it's going to bite someone at some point in some non-obvious way. If this PR didn't have those lines, then I wouldn't have even mentioned it as a comment.

> I agree with keeping unrelated logic separate (per Interface Segregation). In this case, I think the Django storage of opaque keys is a functionality that is closely related to opaque-keys and would need to change if/when we add new KEY_CLASSes, etc. So, keeping it within this repo makes sense (per Single Responsibility).

The only changes opaque-keys had in all of 2017 was a line adding license info to `setup.py` and me adding and then rolling back optimizations I tried to work into it. It's pretty stable at this point. We haven't added a new key class to opaque keys since the CCX keys came in nearly three years ago. Furthermore, when the CCX keys were created, they a) were built in a separate repo by a third party, using the plugin mechanism; and b) subclassed existing key types. They wouldn't require any changes to the Django ORM serialization code, and the code as it stands is still unaware of them. I think you could actually make a plausible argument that Single Responsibility supports keeping them separated -- the keys are stable, but the Django specific serialization code is changing as we transition between 1.8 and 1.11 and beyond.

> I'd like to go forward thinking about the placement of code based on its responsibilities and domain ownership rather than its technology dependencies and constraints. I'm concerned that our current technology-based division of relatively hard to navigate (lib, djangoapps, djangolib, etc) folders within the monolith will transfer to across repos - which can grow into an even larger scale problem. - unless we begin to think about domains/concepts/ownership/responsibilities.

Okay, so stepping away from this PR and talking about the higher level issue -- I am fully supportive of reorganizing the code by domain and accepting that Django is everywhere in edx-platform instead of trying to break things out into lib and such. At the same time, I think there are limits to how much we should ignore our technology dependencies and constraints when separating directories and repos. Security/sandboxing, resource consumption, complexity, language/framework conventions, data source, deployment requirements, rate of change, etc. all factor into this.

As a trivial example, on the FedX side, we want to get to a place where you make a change in any frontend asset and you can see the result in development basically instantly (< 1 second). There are currently a number of configuration values that are dynamically modified in edx-platform's Django settings load, that the JavaScript tooling then queries. The question of whether that belongs in edx-platform is almost immaterial -- we chose to do the front-end processing in JS because the tooling there is vastly superior, and we can't give a 1 second turnaround if we have to invoke something that takes 7 seconds to start up due to framework issues beyond our practical control. We ideally want to be able to process assets on a file-watching docker container that doesn't have Python and the edx-platform dependency soup installed. We should obviously still think hard about domain responsibilities (and I'm trying to make clean separations there), but we can't ignore the technology constraints either.

Again, I'm not trying to argue this particular PR. I wouldn't have approved it if I didn't think it was an acceptable solution. I just think that there are tradeoffs either way, and that these boundaries aren't always going to be clear cut.

robrap:
This approach seems fine to me.  

However, I think removing the try/except is equivalent to logging a message like "This will never happen.", or "If this ever happens, I assume you'll know exactly why and how to fix it."   I prefer keeping the try/catch and logging a message that tells someone why it is happening and how to correct the issue.

cpennington:
I agree that we should keep the `try/catch`, and also that substituting `object` is the wrong thing to do in the `except` clause. If code is trying to import the django-specific key code, it's going to be surprised if those objects don't exist. So, we should fail hard at the import, but with a specific message saying what needs to happen (that the users should `pip install opaque-keys[django]`

nedbat:
On the nitty-gritty of this pattern: I notice that the ImportError's are also marked with "pragma: no cover", which indicates that we are not testing the code in the "no-Django" scenario.  That seems bad to me.

iloveagent57:
Sigh, it would have been easier and simpler to just include Django as a real dependency.  So the main takeaways I'm hearing are:
* We can keep the `try/except` blocks around Django imports.
* The Django-related ImportErrors should result in a hard failure with a useful message (I think I tried this originally, but ran into test configuration issues with the files still being interpretted in the no-Django situation.  But I guess I can probably solve that easily with a good "ignore this directory" command in test configuration).
* The no-Django testing scenario should test that Django imports raise exceptions (and log a message).

nasthagiri:
Sounds good. Thanks @iloveagent57

nedbat:
@iloveagent57 Definitely it is a good thing that we didn't just include Django as a pre-req.  My hackathon project analyzed courses across Open edX, and I used edx-opaque-keys to isolate myself from syntactic details. That is exactly what this library is for: broad use by anyone interested in course ids.  I have no need for Django in my project.

PR #93, opened 2018-03-22
Be explicit about what packages we install

nedbat:
This repo was accidentally installing a "settings" package.  It shouldn't.


nedbat:
@cpennington You agree?

nedbat:
Wait, not right yet.

nedbat:
I'm not including any "tests" subdirectories.

PR #100, opened 2018-12-11
Proposed usability improvements for django key fields

bradenmacdonald:
Two proposed changes to make `UsageKeyField` and friends more useful:

1. Give each of `UsageKeyField` and friends a sensible `max_length` default to DRY up model definitions. Currently we put `max_length=255` in every single model definition where these are used. In addition, use a shorter default length which is compatible with InnoDB indexing when the `utf8mb4` character set is used regardless of the MySQL index length configuration. IMHO any new IDAs we make should be using `utf8mb4` for their entire DB, so the default length should be compatible.
1. It's currently a common pattern to have both a `course_id` field and a `usage_id` field in order to get indexed lookups by course. However, for almost four years we've been using usage keys which have the course ID at the beginning of the serialized usage key value, so it's possible to use only a single indexed `usage_id` column and still query by course using `WHERE usage_key LIKE 'block-v1:COURSE_ID_HERE%'` just as efficiently as if course_id was materialized in its own column and index. With this change, you can create a model that only has a usage_id field and then query by course using the `Model.objects.filter(usage_key__course=self.course_key)` lookup.

Tradeoffs of #2 are:
* Simplifies data model
* Reduces table/index size somewhat by removing redundant data
* Improves data integrity (derived course_id will always correspond to usage_id) - we have observed such a column mismatch bug in the wild on at least one occasion
* But it doesn't support old mongo keys
* It places the constraint on new UsageKey formats that they must encode the course ID in the usage key serialized form in order to be compatible.
* Doesn't support `usage_id__course__in=[list of course IDs]` though that could be added later.
* Prevents `GROUP BY course_id` clauses

**Manual testing**:
Install in your LMS devstack container, then run something like the following from the LMS django shell:

```python
from courseware.models import StudentModule
obj = StudentModule.objects.all()[10]
course_key = obj.course_id
StudentModule.objects.filter(module_state_key__course=course_key)
```

### DIFF
--- a/opaque_keys/edx/django/models.py
+++ b/opaque_keys/edx/django/models.py
@@ -8,7 +8,7 @@ import warnings
 try:
     from django.core.exceptions import ValidationError
     from django.db.models import CharField
-    from django.db.models.lookups import IsNull
+    from django.db.models.lookups import IsNull, StartsWith
 except ImportError:  # pragma: no cover
     # Django is unavailable, none of the classes below will work,
     # but we don't want the class definition to fail when interpreted.
@@ -93,10 +93,13 @@ class OpaqueKeyField(CreatorMixin, CharField):
 
     Empty = object()
     KEY_CLASS = None
+    DEFAULT_MAX_LENGTH = None
 
     def __init__(self, *args, **kwargs):
         if self.KEY_CLASS is None:
             raise ValueError('Must specify KEY_CLASS in OpaqueKeyField subclasses')
+        if self.DEFAULT_MAX_LENGTH is not None:
+            kwargs.setdefault('max_length', self.DEFAULT_MAX_LENGTH)
 
         super(OpaqueKeyField, self).__init__(*args, **kwargs)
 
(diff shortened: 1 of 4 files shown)


opaque_keys/edx/django/models.py
```diff
@@ -190,6 +198,53 @@ class UsageKeyField(OpaqueKeyField):
...
+    """
+    Allows efficiently querying the course that a UsageKey belongs to,
+    without requiring that the CourseKey is stored in a separate column.
+
+    This only works with UsageKeys that contain the full CourseKey in
+    their serialized form, preceding any usage-specific identifiers.
+
+    Usage: my_model.objects.filter(usage_key__course=course_key)
+    """
+    lookup_name = 'course'
```
symbolist: Nit: Since the property name on the UsageKey is `course_key` would that be a nicer name?

opaque_keys/edx/django/models.py
```diff
@@ -190,6 +198,53 @@ class UsageKeyField(OpaqueKeyField):
...
     description = "A Location object, saved to the DB in the form of a string"
     KEY_CLASS = UsageKey
+    # Default to 191 characters as the maximum length of this field type,
+    # because we want to support the MySQL "utf8mb4" character set (real UTF-8)
+    # and the InnoDB index limit on common MySQL configurations is 767 bytes
+    # (191 characters * 4 bytes/character = 764 bytes)
+    DEFAULT_MAX_LENGTH = 191
+
+
+class UsageKeyCourseLookup(StartsWith):
```
symbolist: Nit: Seems a bit out of place here. Maybe in `lookups.py`?

bradenmacdonald:
Going to close this as I haven't felt a compelling need for it lately.

edx-webhook:
Although this pull request is already closed, I've created [OSPR-5465](https://openedx.atlassian.net/browse/OSPR-5465) so that we can track it in Jira.

There is nothing you have to do. No action is needed from your side.  Thanks.

PR #103, opened 2019-04-30
Drop to_dreprecated_string and from_deprecated_string

mikix:
These are old deprecated methods (with replacements 'unicode' and
'from_string' respectively). We don't believe any other code in
the Open edX platform is using them anymore.

DEPR-16


nasthagiri:
Can we update the major version of this library, so clients can realize this is a breaking change.

mikix:
Yeah, I was thinking of doing a separate release PR after this lands.

PR #104, opened 2019-05-15
Declare OEP-7 compliance

jmbowman:
* Test Python 2.7 in Travis again
* Switch from testing Python 3.5 to 3.6
* Drop support for testing against Django 1.8
* Update requirements
* Fix new pylint warnings
* Declare OEP-7 support
* Update ownership metadata


jmbowman:
Closing and reopening to trigger a Travis CI build (it wasn't configured correctly).

openedx.yaml
```diff
@@ -2,6 +2,9 @@
 # http://open-edx-proposals.readthedocs.io/en/latest/oeps/oep-0002.html#specification
 
 nick: opky
-oeps: {}
-owner: edx/platform-team
+oeps:
+    oep-7: true
```
feanil: also, oep-2 and oep-18 compliant?

opaque_keys/edx/locations.py
```diff
@@ -209,9 +209,9 @@ class DeprecatedLocation(BlockUsageLocator):
     The short-lived location:org+course+run+block_type+block_id syntax
     """
     CANONICAL_NAMESPACE = 'location'
-    URL_RE_SOURCE = r"""
-        (?P<org>{ALLOWED_ID_CHARS}+)\+(?P<course>{ALLOWED_ID_CHARS}+)\+(?P<run>{ALLOWED_ID_CHARS}+)\+
-        (?P<block_type>{ALLOWED_ID_CHARS}+)\+
+    URL_RE_SOURCE = u"""
```
feanil: There are backslashes in the ALLOWED_ID_CHARS var, do those need to be updated as well?

jmbowman: No, the `r` at the start of a string literal only affects how characters shown explicitly in the source code for that string literal are treated.  (The combined `ur` prefix no longer exists in Python 3, hence why we need the escaping here).

feanil: ok, so the contents of the var will be escaped correctly before the strings are combined?

jmbowman: Yes, its content is independent of the prefixes used for the string literal referencing it.

PR #107, opened 2019-06-07
ccx uses the tests packages for Mixins.

dianakhuang:
Unfortunately, we use test mixins for https://github.com/edx/ccx-keys , so preventing bundling causes us to not be able to run those tests. Alternate solutions welcome.


nedbat:
I don't understand what you are trying to do, or what the problem is.  Face-to-face?

PR #114, opened 2019-11-15
Removed pylint deprecation warnings

mraarif:
Fixing pylint warnings.


mraarif:
@jmbowman, there's one deprecation warning left (opaque_keys/edx/tests/test_library_usage_locators.py:63) and it's about deprecated `run` property, I've updated other places to use suggested properties but not sure which property to replace `run` with.

mraarif:
or do we want to keep disabling the `pylint` warnings instead?

jmbowman:
Most of these warnings are coming from code that's been deprecated in the `edx-opaque-keys` package, but not yet removed because other edX code may still be using it.  Maybe the best solution for now is to use `# pylint: disable=` comments to disable these specific warnings?  We still want the tests to cover the deprecated code until it's actually removed, but we also want to start getting deprecation warnings from Django and other external packages.

mraarif:
@jmbowman there are some tests that fail if we fix the deprecation warnings e.g in `opaque_keys/edx/tests/test_deprecated_locations.py:50`, are these necessary?

jmbowman:
Those tests are to verify that code still using the deprecated features won't be broken.  I think we just have to live with those warnings until the deprecated code is eventually removed.  I'll create a DEPR ticket to get that removal process started.

mraarif:
closing the PR as discussed.

PR #118, opened 2020-03-19
Fix setup.py to use requirements dir; drop max version constraints

timmc-edx:
Note that some constraints previously had a trailing single-quote, although it doesn't seem to have caused any harm.

This extracts the requirements fix from https://github.com/edx/opaque-keys/pull/116 but does not drop Python 2.7.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,8 +1,41 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+"""
+Package metadata for edx-opaque-keys.
+"""
+from io import open
+
 from setuptools import setup, find_packages
+
+def load_requirements(*requirements_paths):
+    """
+    Load all requirements from the specified requirements files.
+
+    Returns:
+        list: Requirements file relative path strings
+    """
+    requirements = set()
+    for path in requirements_paths:
+        with open(path) as reqs:
+            requirements.update(
+                line.split('#')[0].strip() for line in reqs
+                if is_requirement(line.strip())
+            )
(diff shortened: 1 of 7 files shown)


setup.py
```diff
@@ -18,14 +51,7 @@
...
     packages=find_packages(),
     license='AGPL-3.0',
-    install_requires=[
-        'six>=1.10.0,<2.0.0',
-        'stevedore>=0.14.1,<2.0.0',
-        'pymongo>=2.7.2,<4.0.0'
-    ],
-    extras_require={
-        'django': ['Django>=1.11,<2.0;python_version<"3"', 'Django>=1.11;python_version>"3"']
-    },
```
robrap: @timmc-edx: Did you ever find out why this was pulled out as `extra_requires`, and whether or not that matters?

timmc-edx: No, I can't figure out why it was like that, although I'm still really hazy on how extras work in Python dep management. And it looks like Django never made it into the requirements files, so maybe this needs another (small) PR.

It looks like maybe this is what we want: https://github.com/edx/mockprock/blob/181d2c962006ffcecfda009ee9b6f417982c48c6/setup.py#L61

robrap: What you found looks good if you want to keep the same.  The commit that added it is https://github.com/edx/opaque-keys/commit/c056e200fd8c21819a31d1f6025290521073b2da, and Alex still works here.  So, you could either ask Alex or members of BOM that know more. :)

PR #183, opened 2021-05-24
chore: Bump version to 2.2.1 after upgrading dependencies.

doctoryes:
All Python dependencies were upgraded in this PR: https://github.com/edx/opaque-keys/pull/182

This PR bumps the patch version of the package.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,3 +1,7 @@
+# 2.2.1
+
+* Upgraded dependencies
+
 # 2.2.0
 
 * Dropped support for Python3.5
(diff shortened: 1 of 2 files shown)


kdmccormick (approved):

UsamaSadiq (approved):
The requirements upgrade job is scheduled to run on weekly basis. Do we want to bump the version each time the requirements dependencies are upgraded?

doctoryes:
@UsamaSadiq If we had an automated process to do so, I'd say yes. Since we'd currently be doing the version bump manually, I don't think we need to do the version bump *every* time - but doing it semi-regularly would be healthy. Releasing often promotes quality via smaller, more increment releases.

doctoryes:
@stvstnfrd FYI: I've just bumped the version and released opaque-keys to release dependency upgrades. I was told that releases of this pkg had caused Stanford some issues in the past - so wanted to let you know.

PR #190, opened 2021-07-08
docs: Build the docs as a part of CI.

feanil:
We do this so that we can catch any issues with the RSTs before they
cause build failures in the final build.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,0 +1,21 @@
+# .readthedocs.yaml
+# Read the Docs configuration file
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
+version: 2
+
+# Build documentation in the docs/ directory with Sphinx
+sphinx:
+   configuration: docs/conf.py
+
+# Optionally build your docs in additional formats such as PDF
+formats:
+   - pdf
+   - epub
+
+# Optionally set the version of Python and requirements required to build your docs
+python:
+   version: 3.8
+   install:
+   - requirements: requirements/doc.txt
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ python:
   - '3.8'
 env:
   - TOXENV=quality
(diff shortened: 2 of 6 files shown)


nedbat:
I tried this PR, and added mistakes into the .rst files.  `tox -e docs` showed warnings, and then succeeded:
```
...
docs run-test: commands[0] | make -C docs html
sphinx-build -b html -d _build/doctrees   . _build/html
Running Sphinx v4.0.2

...

/src/opaque-keys/docs/index.rst:11: WARNING: Inline interpreted text or phrase reference start-string without end-string.
/src/edx/src/opaque-keys/docs/index.rst:13: WARNING: Problems with "include" directive path:
InputError: [Errno 2] No such file or directory: 'foo.bar'.
/src/edx/src/opaque-keys/docs/index.rst:9: WARNING: Unknown target name: "foo".
/src/edx/src/opaque-keys/.tox/docs/lib/python3.8/site-packages/opaque_keys/edx/keys.py:docstring of opaque_keys.edx.keys.UsageKeyV2:9: WARNING: Unexpected indentation.
/src/edx/src/opaque-keys/.tox/docs/lib/python3.8/site-packages/opaque_keys/edx/locator.py:docstring of opaque_keys.edx.locator.BlockUsageLocator:19: WARNING: Unexpected indentation.

...
The HTML pages are in _build/html.

Build finished. The HTML pages are in _build/html.
___________________________________________ summary ________________________________________
  docs: commands succeeded
  congratulations :)
```
(Note the last two warnings were pre-existing.)

So if we want things to fail, we may have too lenient a chain here.  I think the flags on the sphinx-build command need to change.  Adding the -W switch somehow (edit the Makefile, provide it as an environment variable, etc), will force things to fail if there's a problem.   The pre-existing warnings are about the contents of docstrings, so those would have to be fixed.

See https://github.com/edx/opaque-keys/pull/191

nedbat (approved):

PR #191, opened 2021-07-13
docs: make the docs strict, and fix two docstrings

nedbat:
The -W flag will turn all warnings into errors, so the build will fail
if there are problems.

Three existing problems are fixed.


nedbat:
Equivalent changes were made in #190.

PR #206, opened 2021-12-09
chore: pin pymongo version

matthugs:


requirements/constraints.txt
```diff
@@ -21,3 +21,6 @@ pytest-xdist<2.0.0
 # include pytest) doesn't generate `ci.txt` with a pluggy version >=1. 
 # Remove this pin when pytest no longer requires pluggy<1.
 pluggy<1
+
+# migration guide here: https://pymongo.readthedocs.io/en/4.0/migrate-to-pymongo4.html
+pymongo<4.0
```
doctoryes, line 26: There's an existing pymongo constraint in `base.in`. It doesn't conflict with this new one - but the two should be consolidated.

Also, adding this constraint to the `constraints.txt` file means that the constraint *won't* be included in the `install_requires` of the `setup.py` file - by my reading of the code in `setup.py`.  So shouldn't this constraint be left in the `base.in` file for that reason?

matthugs: I believe the new setup.py in https://github.com/edx/opaque-keys/pull/202/files#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7R53 does take the contents of `constraints.txt` into account; it just doesn't add the package if it wasn't already present in `base.in`'s required packages. This is in-line with the different semantics between including a requirements file via `-r` versus `-c`.

I would prefer to keep the lines separate because to me they intend to do different things, but I'm easy.

doctoryes: I see - I didn't realize this pattern existed in other edX-owned packages where there's a separate "the app needs this range of version" (`base.in`) from "no app should use above this version" (`constraints.txt`). I've always tended to consolidate a package's constraints so that no one will look at one of the two constraints and think that only that constraint exists.

If this is a pattern already adopted in other repos, I'm fine with having the lines separate. I'll approve and leave it up to you.

doctoryes (approved):

PR #246, opened 2022-08-12
fix: correct minor code stule issue, so python requirements update will work. No ticket associated.

kenclary:


connorhaugh (approved):
:shipit:

PR #256, opened 2023-07-21
feat: Add type hints

bradenmacdonald:
This adds type hints to opaque-keys, including to the CI build so that type rules are enforced before PRs can be merged.

This is actually aimed to help with edx-platform development, because once these type hints are in place, mypy can be used to check types of modules in edx-platform - see https://github.com/openedx/edx-platform/pull/32591 for example. Currently mypy just treats any OpaqueKey type as `Any` which is not helpful.

Some of the variables are of type `ObjectId` but this gets broadened automatically to `Any` because we're using an old version of `pymongo` that lacks type annotations. Once we upgrade `pymongo`, we'll have a more specific `ObjectId` type working.

Left for future:
- [ ] Upgrade `pymongo` so we have a working `ObjectId` type

PR #259, opened 2023-08-24
fix: type hints weren't installed properly, add type hints for Django Fields

bradenmacdonald:
My last PR #256 wasn't *quite* working with `mypy` in edx-platform. It was working fine in "editable" mode with `python setup.py develop` or `pip install -e .` but when installed via `pip install edx-opaque-keys==2.5.0` or `python setup.py install`, the important `py.typed` marker file was missing.

I thought that using `include_package_data=True` was sufficient, but it turns out I also needed to add a line to `MANIFEST.in`.

With this change, the `py.typed` file is installed and mypy will use the type hints added in #256.

------

I also added type hints for the django fields like `UsageKeyField`. This only works because of the django-stubs mypy plugin that we use in edx-platform, and requires setting a "private" class attribute to define the types, but it seems to work very well. Now mypy is aware that reading from a `UsageKeyField` will result in a `UsageKey` and writing to one accepts a `UsageKey` or `str` or `None`.  (Keep in mind these additional static fields are type-only and will be ignored at runtime; they only affect mypy and then only if the django-stubs mypy plugin is installed.)

### DIFF
--- a/opaque_keys/edx/django/models.py
+++ b/opaque_keys/edx/django/models.py
@@ -194,6 +194,12 @@ class LearningContextKeyField(OpaqueKeyField):
     """
     description = "A LearningContextKey object, saved to the DB in the form of a string"
     KEY_CLASS = LearningContextKey
+    # Declare the field types for the django-stubs mypy type hint plugin.
+    # See https://github.com/typeddjango/django-stubs/blob/6a850f6/django-stubs/db/models/fields/__init__.pyi#L503-L511
+    # for examples of how this works for the standard field types in the upstream codebase.
+    # Note that these particular type annotations have no effect at runtime nor on Django itself nor on PyLance.
+    _pyi_private_set_type: LearningContextKey | str | None  # The types that you can set into this field.
+    _pyi_private_get_type: LearningContextKey  # The type that you get when you read from this field
 
 
 class CourseKeyField(OpaqueKeyField):
@@ -202,6 +208,9 @@ class CourseKeyField(OpaqueKeyField):
     """
     description = "A CourseKey object, saved to the DB in the form of a string"
     KEY_CLASS = CourseKey
+    # Declare the field types for the django-stubs mypy type hint plugin:
+    _pyi_private_set_type: CourseKey | str | None
+    _pyi_private_get_type: CourseKey
 
 
 class UsageKeyField(OpaqueKeyField):
@@ -210,6 +219,9 @@ class UsageKeyField(OpaqueKeyField):
     """
     description = "A Location object, saved to the DB in the form of a string"
     KEY_CLASS = UsageKey
+    # Declare the field types for the django-stubs mypy type hint plugin:
+    _pyi_private_set_type: UsageKey | str | None
+    _pyi_private_get_type: UsageKey
 
 
 class LocationKeyField(UsageKeyField):
@@ -228,3 +240,6 @@ class BlockTypeKeyField(OpaqueKeyField):
     """
     description = "A BlockTypeKey object, saved to the DB in the form of a string."
     KEY_CLASS = BlockTypeKey
+    # Declare the field types for the django-stubs mypy type hint plugin:
+    _pyi_private_set_type: BlockTypeKey | str | None
+    _pyi_private_get_type: BlockTypeKey
--- a/opaque_keys/edx/django/models.py
+++ b/opaque_keys/edx/django/models.py
@@ -199,7 +199,7 @@ class LearningContextKeyField(OpaqueKeyField):
     # for examples of how this works for the standard field types in the upstream codebase.
     # Note that these particular type annotations have no effect at runtime nor on Django itself nor on PyLance.
     _pyi_private_set_type: LearningContextKey | str | None  # The types that you can set into this field.
-    _pyi_private_get_type: LearningContextKey  # The type that you get when you read from this field
+    _pyi_private_get_type: LearningContextKey | None  # The type that you get when you read from this field
 
 
 class CourseKeyField(OpaqueKeyField):
@@ -210,7 +210,7 @@ class CourseKeyField(OpaqueKeyField):
     KEY_CLASS = CourseKey
     # Declare the field types for the django-stubs mypy type hint plugin:
     _pyi_private_set_type: CourseKey | str | None
-    _pyi_private_get_type: CourseKey
+    _pyi_private_get_type: CourseKey | None
 
 
 class UsageKeyField(OpaqueKeyField):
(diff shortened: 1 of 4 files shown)


openedx-webhooks:
Thanks for the pull request, @bradenmacdonald!  Please note that it may take us up to several weeks or months to complete a review and merge your PR.

Feel free to add as much of the following information to the ticket as you can:
- supporting documentation
- [Open edX discussion forum threads](https://discuss.openedx.org/)
- timeline information ("this must be merged by XX date", and why that is)
- partner information ("this is a course on edx.org")
- any other information that can help Product understand the context for the PR

All technical communication about the code itself will be done via the GitHub pull request interface. As a reminder, [our process documentation is here](http://edx-developer-guide.readthedocs.org/en/latest/process/overview.html).


Please let us know once your PR is ready for our review and all tests are green.

mphilbrick211:
Hi @openedx/teaching-and-learning! Would someone please be able to review/merge this for us?

bradenmacdonald:
Yes, thanks - I would appreciate a review of this one-line change. But I will merge it myself as I've realized I need to merge some small type fixes to edx-platform first.

mariajgrimaldi (commented):
I'm reviewing this since it was mentioned during the last CC meeting. I hope it helps! 

I installed this version in a local venv:

```
(venv) ~/venv/lib/python3.8/site-packages/opaque_keys ls
edx  __init__.py  __pycache__  py.typed  tests
```

While in my current nightly installation with edx-opaque-keys 2.5.0:

```
app@d3bd25e0d8c4:~/venv/lib/python3.8/site-packages/opaque_keys$ ls
edx  __init__.py  __pycache__  tests
```

Then, I removed [this comment](https://github.com/openedx/edx-platform/pull/33104/files#diff-e7682ca30533ea4960653ff3f4516608f7a57c9cfb6518489ae77f14906a2369R66) just so type-checking warnings are raised. Then, I ran mypy with edx-opaque-keys==2.5.0 in edx-platform and got:
```
app@d3bd25e0d8c4:~/edx-platform$ mypy
Success: no issues found in 103 source files
```

After installing this version I got:

app@d3bd25e0d8c4:~/edx-platform$ mypy
```
openedx/core/djangoapps/content_staging/data.py:66: error: Only concrete class can be given where "Type[UsageKey]" is expected  [type-abstract]
Found 1 error in 1 file (checked 103 source files)
```
So now it's working! Thanks

opaque_keys/edx/django/models.py
```diff
@@ -194,6 +194,12 @@ class LearningContextKeyField(OpaqueKeyField):
     """
     description = "A LearningContextKey object, saved to the DB in the form of a string"
     KEY_CLASS = LearningContextKey
+    # Declare the field types for the django-stubs mypy type hint plugin.
+    # See https://github.com/typeddjango/django-stubs/blob/6a850f6/django-stubs/db/models/fields/__init__.pyi#L503-L511
+    # for examples of how this works for the standard field types in the upstream codebase.
+    # Note that these particular type annotations have no effect at runtime nor on Django itself nor on PyLance.
+    _pyi_private_set_type: LearningContextKey | str | None  # The types that you can set into this field.
+    _pyi_private_get_type: LearningContextKey  # The type that you get when you read from this field
```
ormsbee, line 202: So I've seen this pattern elsewhere, and I totally believe you're doing the right thing. But I don't understand this part–why is `None` a settable type, but not a gettable type for this field?

antoviaque:
@bradenmacdonald FYI your call for review has been discussed during the [contributors meetup this week](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3856924673/2023-09-05+CC+Working+Group+Meeting+Notes) - there are potentially 3 reviewers: @mariajgrimaldi will have a look, Stefania will ask @ghassanmas and @auraz will try to get someone from Raccoon Gang.

opaque_keys/edx/django/models.py
bradenmacdonald: @ormsbee 

> why is None a settable type, but not a gettable type for this field?

My understanding was that for these type definitions you generally don't specify `None` as either a setter or a getter, and the mypy plugin will automatically add the `| None` union type if you declare a specific field instance with `null=True`. And I noticed that even if you don't declare it with `null=True`, you can still set a `None` value because the field type [converts](https://github.com/openedx/opaque-keys/blob/e0ee525df4395f3795501fdc3a75a7ecdbedcb1f/opaque_keys/edx/django/models.py#L131-L133) that to an empty string.

That said, I think I was not really thinking things through. `OpaqueKeyField`s should never be declared as `null=True`, because they're `CharField`-based, but instead they should use `blank=True` or not to indicate whether they accept empty values. The problem is that `OpaqueKeyField` will convert empty strings to `None` in the getter, which mypy is unaware of without `null=True`. So in the end, I think it's best to just explicitly specify `None` as one of the possible types on both the getter and the setter, which I have now done.

Thanks for the helpful question. I have pushed a commit to add `| None` to the getters.

bradenmacdonald:
Thanks for testing this @mariajgrimaldi!

I still technically need an approval from someone with merge rights on this specific repo before I can merge, but it should be easier to find with your test done.

antoviaque:
@itsjeyd See @bradenmacdonald 's request above ^ Do we have a list of people who can give a :+1: on this repo? I see from https://openedx.atlassian.net/wiki/spaces/COMM/pages/3632070677/Pilot+Phase+3 that it's being taken on by Axim for maintainership in phase 3, though that's likely not the only ones who can give a binding :+1: ?

And @bradenmacdonald any suggestion who would be a good person for the binding review, even if that person doesn't have access to it yet?

bradenmacdonald:
@felipemontoya would you be able and comfortable approving this PR like you did the last one? :)

> Do we have a list of people who can give a 👍 on this repo?

@antoviaque Isn't it as simple as [this list](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3156344833/Core+Contributors+to+the+Open+edX+Project)? There are three people, but one is me. I probably should have just pinged them directly [as I did last time](https://github.com/openedx/opaque-keys/pull/256#issuecomment-1650194125) since that eventually worked.

ormsbee (approved):

openedx-webhooks:
@bradenmacdonald 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/opaque-keys&entry.867055334=https://github.com/openedx/opaque-keys/pull/259&entry.1484655318=https://github.com/bradenmacdonald&entry.752974735=2023-08-24+19:43&entry.1917517419=2023-09-21+18:29&entry.2133058324=Yes) so we can improve your experience in the future.

bradenmacdonald:
@felipemontoya @antoviaque Never mind - as you can see now that @ormsbee is back I got his help :) Thx anyways!

itsjeyd:
@antoviaque @bradenmacdonald Glad to see that you already found a reviewer :)

auraz:
Hey @antoviaque JFYI I'm not in Raccoon Gang for 4+ years.

PR #317, opened 2024-06-07
feanil/unpin pymongo

feanil:
- **fix: Unpin Pymongo.**
  - We only use very simple objects from pymongo in this repo and none of
    them have changes in the intermediate versions.  The only usage of the
    SON object is to instantiate one and so no code in this repo is going to
    break by this upgrade.

    There may be other repos that upstream that might need to pin pymongo
    based on their usage but it shouldn't be pinned here.
- **chore: Run `make upgrade`**

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,3 +1,7 @@
+# 2.10.0
+
+* Unpin pymongo and upgrade to the latest available version.
+
 # 2.9.0
 
 * Upgrade pymongo to 4.4. It contains few breaking changes.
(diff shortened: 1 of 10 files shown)


feanil:
@ormsbee @kdmccormick tagging you in case there is some historic context I have for why this pin is here.

farhan (approved):

kdmccormick (approved):
Here's the PR that pinned it: https://github.com/openedx/opaque-keys/pull/304

It seems like it was pinned out of an abundance of caution over this breaking change: https://pymongo.readthedocs.io/en/4.4.0/migrate-to-pymongo4.html#son-items-now-returns-dict-items-object

On one hand, these BSON objects are part of the opaque-keys interface, so breaking changes to BSON could be considered breaking changes to opaque-keys.

On the other hand, the breakage is very minor, and also very easy to detect: there'll be a ValueError, TypeError or AttributeError if someone was depending on `.items()` to return a list. It's not like some pernicious encoding change that could fly under the radar in downstream packages. So, I think it's right to unpin, and let downstreams worry (or not) about the breaking pymongo change.

PR #335, opened 2024-08-27
feat: LibraryCollectionKey and LibraryCollectionLocator created [FC-0062]

ChrisChV:
## Description

Creates `LibCollectionKey` and `LibCollectionLocator`

## Supporting information

- Github Issue: https://github.com/openedx/modular-learning/issues/228
- This change is used on: https://github.com/openedx/edx-platform/pull/35383
- Internal ticket: https://tasks.opencraft.com/browse/FAL-3785

## Testing instructions

- Check all the code
- Verify that the test covers the changes

### DIFF
--- a/opaque_keys/edx/keys.py
+++ b/opaque_keys/edx/keys.py
@@ -89,6 +89,36 @@ class CourseKey(LearningContextKey):
         raise NotImplementedError()
 
 
+class CollectionKey(LearningContextKey):
+    """
+    An :class:`opaque_keys.OpaqueKey` identifying a particular Library Collection object.
+    """
+    __slots__ = ()
+
+    @property
+    @abstractmethod
+    def org(self) -> str | None:  # pragma: no cover
+        """
+        The organization that this collection belongs to.
+        """
+        raise NotImplementedError()
+
+    @property
+    @abstractmethod
+    def lib(self) -> str | None:  # pragma: no cover
+        """
+        The lib for this collection.
+        """
+        raise NotImplementedError()
+
+    @property
+    def context_key(self) -> LearningContextKey:
+        """
+        Get the learning context key (LearningContextKey) for this XBlock usage.
+        """
+        raise NotImplementedError()
+
+
 class DefinitionKey(OpaqueKey):
     """
     An :class:`opaque_keys.OpaqueKey` identifying an XBlock definition.
--- a/opaque_keys/edx/keys.py
+++ b/opaque_keys/edx/keys.py
@@ -98,19 +98,14 @@ class LibraryCollectionKey(OpaqueKey):
     An :class:`opaque_keys.OpaqueKey` identifying a particular Library Collection object.
     """
     KEY_TYPE = 'collection_key'
+    library_key: LibraryLocatorV2
+    collection_id: str
     __slots__ = ()
 
     @property
     def org(self) -> str | None:  # pragma: no cover
         """
         The organization that this collection belongs to.
-        """
-        raise NotImplementedError()
-
-    @property
-    def library_key(self) -> LibraryLocatorV2:  # pragma: no cover
-        """
-        Get the key of the library that this collection belongs to.
         """
         raise NotImplementedError()
 
--- a/opaque_keys/edx/keys.py
+++ b/opaque_keys/edx/keys.py
@@ -112,7 +112,7 @@ class LibCollectionKey(LearningContextKey):
         raise NotImplementedError()
 
     @property
-    def context_key(self) -> LearningContextKey:
+    def context_key(self) -> LearningContextKey:  # pragma: no cover
         """
         Get the learning context key (LearningContextKey) for this XBlock usage.
         """
--- a/opaque_keys/edx/keys.py
+++ b/opaque_keys/edx/keys.py
@@ -89,10 +89,11 @@ class CourseKey(LearningContextKey):
         raise NotImplementedError()
 
 
-class LibCollectionKey(LearningContextKey):
+class LibraryCollectionKey(OpaqueKey):
     """
     An :class:`opaque_keys.OpaqueKey` identifying a particular Library Collection object.
     """
+    KEY_TYPE = 'collection_key'
     __slots__ = ()
 
     @property
@@ -112,9 +113,9 @@ class LibCollectionKey(LearningContextKey):
         raise NotImplementedError()
 
     @property
-    def context_key(self) -> LearningContextKey:  # pragma: no cover
+    def library_key(self) -> LearningContextKey:  # pragma: no cover
         """
-        Get the learning context key (LearningContextKey) for this XBlock usage.
+        Get the key of the library that this collection belongs to.
         """
         raise NotImplementedError()
 
--- a/opaque_keys/edx/keys.py
+++ b/opaque_keys/edx/keys.py
@@ -4,11 +4,15 @@ OpaqueKey abstract classes for edx-platform object types (courses, definitions, 
 from __future__ import annotations
 import json
 from abc import abstractmethod
+from typing import TYPE_CHECKING
 import warnings
 
 from typing_extensions import Self  # For python 3.11 plus, can just use "from typing import Self"
 
 from opaque_keys import OpaqueKey
+
+if TYPE_CHECKING:
+    from opaque_keys.edx.locator import LibraryLocatorV2
 
 
 class LearningContextKey(OpaqueKey):  # pylint: disable=abstract-method
@@ -97,7 +101,6 @@ class LibraryCollectionKey(OpaqueKey):
     __slots__ = ()
 
     @property
-    @abstractmethod
     def org(self) -> str | None:  # pragma: no cover
         """
         The organization that this collection belongs to.
@@ -105,15 +108,7 @@ class LibraryCollectionKey(OpaqueKey):
         raise NotImplementedError()
 
     @property
-    @abstractmethod
-    def lib(self) -> str | None:  # pragma: no cover
-        """
-        The lib for this collection.
-        """
-        raise NotImplementedError()
-
-    @property
-    def library_key(self) -> LearningContextKey:  # pragma: no cover
+    def library_key(self) -> LibraryLocatorV2:  # pragma: no cover
         """
         Get the key of the library that this collection belongs to.
         """
--- a/opaque_keys/edx/keys.py
+++ b/opaque_keys/edx/keys.py
@@ -89,7 +89,7 @@ class CourseKey(LearningContextKey):
         raise NotImplementedError()
 
 
-class CollectionKey(LearningContextKey):
+class LibCollectionKey(LearningContextKey):
     """
     An :class:`opaque_keys.OpaqueKey` identifying a particular Library Collection object.
     """
--- a/opaque_keys/edx/locator.py
+++ b/opaque_keys/edx/locator.py
@@ -15,7 +15,8 @@ from bson.son import SON
 from typing_extensions import Self  # For python 3.11 plus, can just use "from typing import Self"
 
 from opaque_keys import OpaqueKey, InvalidKeyError
-from opaque_keys.edx.keys import AssetKey, CourseKey, DefinitionKey, LearningContextKey, UsageKey, UsageKeyV2
+from opaque_keys.edx.keys import AssetKey, CourseKey, DefinitionKey, \
+    LearningContextKey, UsageKey, UsageKeyV2, CollectionKey
 
 log = logging.getLogger(__name__)
 
@@ -1620,3 +1621,55 @@ class LibraryUsageLocatorV2(CheckFieldMixin, UsageKeyV2):
         # HTML5 allows ID values to contain any characters at all other than spaces.
         # These key types don't allow spaces either, so no transform is needed.
         return str(self)
+
+
+class CollectionLocator(CheckFieldMixin, CollectionKey):
+    """
+    When serialized, these keys look like:
+        lib-collection:org:lib:collection-id
+    """
+    CANONICAL_NAMESPACE = 'lib-collection'
+    KEY_FIELDS = ('org', 'lib', 'usage_id')
+    org: str
+    lib: str
+    usage_id: str
+
+    __slots__ = KEY_FIELDS
+    CHECKED_INIT = False
+
+    # Allow usage IDs to contian unicode characters
+    USAGE_ID_REGEXP = re.compile(r'^[\w\-.]+$', flags=re.UNICODE)
+
+    def __init__(self, org: str, lib: str, usage_id: str):
+        """
+        Construct a CollectionLocator
+        """
+        self._check_key_string_field("org", org)
+        self._check_key_string_field("lib", lib)
+        self._check_key_string_field("usage_id", usage_id, regexp=self.USAGE_ID_REGEXP)
+        super().__init__(
+            org=org,
+            lib=lib,
+            usage_id=usage_id,
+        )
+
+    @property
+    def context_key(self) -> LibraryLocatorV2:
+        return LibraryLocatorV2(self.org, self.lib)
+
+    def _to_string(self) -> str:
+        """
+        Serialize this key as a string
+        """
+        return ":".join((self.org, self.lib, self.usage_id))
+
+    @classmethod
+    def _from_string(cls, serialized: str) -> Self:
+        """
+        Instantiate this key from a serialized string
+        """
+        try:
+            (org, lib, usage_id) = serialized.split(':')
+            return cls(org, lib, usage_id)
+        except (ValueError, TypeError) as error:
+            raise InvalidKeyError(cls, serialized) from error
--- a/opaque_keys/edx/locator.py
+++ b/opaque_keys/edx/locator.py
@@ -1629,8 +1629,8 @@ class LibraryCollectionLocator(CheckFieldMixin, LibraryCollectionKey):
         lib-collection:org:lib:collection-id
     """
     CANONICAL_NAMESPACE = 'lib-collection'
-    KEY_FIELDS = ('lib_key', 'collection_id')
-    lib_key: LibraryLocatorV2
+    KEY_FIELDS = ('library_key', 'collection_id')
+    library_key: LibraryLocatorV2
     collection_id: str
 
     __slots__ = KEY_FIELDS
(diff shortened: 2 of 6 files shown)


openedx-webhooks:
Thanks for the pull request, @ChrisChV!

### What's next?

*Please work through the following steps to get your changes ready for engineering review:*

#### :radio_button: Get product approval

#### :radio_button: Provide context

- Dependencies
- Blockers
- Timeline information
- Partner information
- Supporting documentation





#### :radio_button: Get a green build

#### :radio_button: Let us know that your PR is ready for review:

### Who will review my changes?


This repository is currently maintained by `@openedx/axim-engineering`.  Tag them in a comment and let them know that your changes are ready for review.


### Where can I find more information?

### When can I expect my changes to be merged?

- The need for product review

opaque_keys/edx/locator.py
```diff
@@ -1620,3 +1621,55 @@ def html_id(self) -> str:
         # HTML5 allows ID values to contain any characters at all other than spaces.
         # These key types don't allow spaces either, so no transform is needed.
         return str(self)
+
+
+class CollectionLocator(CheckFieldMixin, CollectionKey):
```
bradenmacdonald, line 1626: If this is specific to libraries, it should be called `LibCollectionLocator`

opaque_keys/edx/keys.py
```diff
@@ -89,6 +89,36 @@ def make_asset_key(self, asset_type: str, path: str) -> AssetKey:  # pragma: no
         raise NotImplementedError()
 
 
+class CollectionKey(LearningContextKey):
```
bradenmacdonald, line 92: If this is specific to libraries, it should be called `LibCollectionKey`

bradenmacdonald:
@ormsbee I remember you had some concerns about introducing a new opaque key. Can you let us know your thoughts?

The tagging API currently assumes that anything we apply tags to has an opaque key, so in order to allow tagging of Collections, we need to either implement this key or remove that assumption.

ormsbee:
> @ormsbee I remember you had some concerns about introducing a new opaque key. Can you let us know your thoughts?

@bradenmacdonald: This is the [Slack thread](https://axim-collaborative.slack.com/archives/C05NT2YN820/p1719000013106399?thread_ts=1718133252.323079&cid=C05NT2YN820) where I asked the question about whether we really needed them (you, @kdmccormick, and @yusuf-musleh commented).

It boils down to the fact that OpaqueKey machinery is weird and complex for what it does (even explaining to someone why there's a `CollectionKey` and a `CollectionLocator` is weird), so do we really need it for this feature?

Also, it's plausible that we'd want to change a library's key at some point, which is more feasible to do now since we're not storing as many references to it. Keys in `openedx_learning` app models are almost all relative to the `LearningPackage` you're in. The only top level one that's stored is the library key on the `LearningPackage` itself (and I'm not sure even that's desirable, since we store it again in edx-platform side).

> The tagging API currently assumes that anything we apply tags to has an opaque key, so in order to allow tagging of Collections, we need to either implement this key or remove that assumption.

I was under the impression that any sort of identifier would do, and you mention in the thread that UUIDs would work as well. Is it important that they be humanly readable? If it's just backend machinery at the moment, I'd prefer to use UUIDs for now. Or even the primary key of the thing being tagged if it's easy–since that would be the most compact representation. I know some other tagging libraries actually make a model-per-thing-being-tagged.

In the longer term, I think that tagging functionality will be simpler and easier to adopt into other systems if it does not force a requirement for OpaqueKeys (e.g. forums posts/comments, catalog information, students, etc.). I would expect that `openedx_tagging` eventually moves into its own, top level repo.

All that being said, I don't know how much work any of this is. If making it more general threatens the Sumac timeline, I'm okay with  rolling forward with a `CollectionKey` for now.

FYI @kdmccormick

bradenmacdonald:
@ormsbee We can of course use UUIDs for the things that are being tagged. I was thinking that if we use primary keys, or perhaps even if we want to use UUIDs, we'd want some kind of namespacing so we have an idea of what sort of thing is being tagged and/or can avoid PK collisions. And if we did that (`lib-collection:23489` as an example), then whatever idiosyncratic namespacing we implement is not that different from opaque keys. So it would be weird to implement a second type of string key namespacing, when we already have a library for it (although the key vs. locator thing is weird, and it would be nice to get rid of it).

However, I guess if we're strictly using UUIDs, then we don't technically need namespacing at all. It may be hard to debug what thing is actually tagged though, since we'll just see a random UUID and have no idea what type of thing it is. Thoughts?

ormsbee:
Oh right, I'm forgetting that there's a single global tagging table and that we need to namespace. Would it be terrible if we stored UUIDs and had a 1:1 join table that matched those rows with `PublishableEntity`? Or two tables, one to match against `Published` and the other `Draft`? (Sorry, I don't remember how we're distinguishing between those two in the tagging.)

bradenmacdonald:
[Here is the current ID field; it's just a string ID](https://github.com/openedx/openedx-learning/blob/baebbb3a1398a151f4fbe917147979419c6a8a45/openedx_tagging/core/tagging/models/base.py#L771-L776). Tagging is not part of the publishing cycle so doesn't distinguish between published and draft things.

We want to associate that `ObjectTag` model with some [`Collection`](https://github.com/openedx/openedx-learning/blob/baebbb3a1398a151f4fbe917147979419c6a8a45/openedx_learning/apps/authoring/collections/models.py#L19) instances, which are themselves not PublishableEntities. So I think we just need some way of putting the `Collection`'s primary key into the `ObjectTag.object_id` string field. We _could_ also add an additional [generic foreign key](https://docs.djangoproject.com/en/5.1/ref/contrib/contenttypes/#generic-relations) to `ObjectTag` but that's getting complicated. Since `Collection` is also in the same overall package as tagging, we could also create a join table specifically for Collection-ObjectTag, but I'm not sure what benefit that would offer.

ormsbee:
@bradenmacdonald: Please pardon the silly questions stemming from my ignorance of the data model, and my brain being stuck on components and not collections.

If the same tag from the same taxonomy applies to three totally different types of things (e.g. a component, a collection, and a forum comment), does the data model have any way to differentiate them? Or does it rely purely on the `object_id` (and consequently OpaqueKeys) to do the dispatch between them?

bradenmacdonald:
@ormsbee 

>  Or does it rely purely on the object_id (and consequently OpaqueKeys) to do the dispatch between them?

That.

> it seems like basic filtering by type would be way harder than it should be?

We haven't had any use case for basic filtering by type, and because it's a string prefix search on an indexed column, it's also extremely efficient at the moment even without a dedicate column.

We could add another column that's a foreign key to the `contenttype` of the thing that's being tagged, certainly - but we'd have to carefully review the code in various places which assumes `object_id` is the entire reference to the thing being tagged.

ormsbee:
Like I said before, I don't want to jeopardize the timeline over this, and I'm not going to block the merging of new key types. As a general rule, I would like more of our data model to be explicitly structured at the database layer, and not rely on Python code to introspect things (or having to know that key type X always starts with one of these two strings). I realize that to a certain extent we _have_ to do that, because so much of our content is still in the ModuleStore and serialized to MongoDB, but I don't want it to be our long term direction.

rpenido:
> We haven't had any use case for basic filtering by type, although it could currently be done as a string prefix search on object_id (an indexed column), which means it's already extremely efficient even without a dedicated column.

Also, having some kind of hierarchy (which the OpaqueKey has) could be handy. If we nuke a Library with all the Components and Collections, we can use the `object_id` to clear all tags without iterating the children. It also makes it easier to find orphaned items.

opaque_keys/edx/locator.py
```diff
@@ -1620,3 +1621,59 @@ def html_id(self) -> str:
...
+    org: str
+    lib: str
+    lib_key: LibraryLocatorV2
+    usage_id: str
+
+    __slots__ = KEY_FIELDS
+    CHECKED_INIT = False
+
+    # Allow usage IDs to contian unicode characters
+    USAGE_ID_REGEXP = re.compile(r'^[\w\-.]+$', flags=re.UNICODE)
```
rpenido, line 1642: This is the collection id (a number), but I'm unsure if we want to make it stricter here.

Maybe for clarity? What do you think?

rpenido (approved):
LGTM :+1:
Thank you for your work, @ChrisChV!

I only have a little to add here as I don't have much context with the OpaqueKeys.

- [x] I tested this using the instructions from https://github.com/openedx/edx-platform/pull/35383
- [x] I read through the code
- [x] ~I checked for accessibility issues~
- [x] Includes documentation

opaque_keys/edx/locator.py
ormsbee: Wait, it's the ID? As in the primary key? We can't use that as part of the key if there's ever an intention to make it importable into a different instance.

bradenmacdonald: At the moment we don't have any other identifier for collections; we'll need to add either a UUID or a slug (or both). If a slug, it only needs to be unique within a library. I prefer slugs for readability. For now the slug can be auto-generated by [slugifying](https://docs.djangoproject.com/en/5.1/ref/utils/#django.utils.text.slugify) the collection name and appending a number until it's unique.

ormsbee: Yeah, I think a slug is fine, as long as we write it to the database–say, a `key` field on `Collection` with a unique constraint on the combo of that and the LearningPackage (like how PublishableEntities work today).

bradenmacdonald: Cool, let's do that ^ @rpenido . Sorry for not realizing the need for a slug sooner. CC @pomegranited

opaque_keys/edx/keys.py
```diff
@@ -89,6 +89,36 @@ def make_asset_key(self, asset_type: str, path: str) -> AssetKey:  # pragma: no
         raise NotImplementedError()
 
 
+class LibCollectionKey(LearningContextKey):
```
kdmccormick, line 92:
```suggestion
class LibraryCollectionKey(OpaqueKey):
```

1. My sense is that a collection is not a learning context. More concretely, I can't imagine us wanting to store learner state against a collection, the way that we store learner state against a CourseKey or (for previewing purposes) a LibraryKey. Would it cause any issue to simply subclass OpaqueKey directly?
2. For the sake of the API, let's be consistent between "Lib" and "Library".  We have `lib:` in the keys, so I like that you chose `lib-collection:`. But we spell out `LibraryLocatorV2`, so I'd prefer if you spelled out `LibraryCollectionKey` and `LibraryCollectionLocator` as well.

opaque_keys/edx/keys.py
```diff
@@ -89,6 +89,36 @@ def make_asset_key(self, asset_type: str, path: str) -> AssetKey:  # pragma: no
...
+        The lib for this collection.
+        """
+        raise NotImplementedError()
+
+    @property
+    def context_key(self) -> LearningContextKey:  # pragma: no cover
+        """
+        Get the learning context key (LearningContextKey) for this XBlock usage.
+        """
+        raise NotImplementedError()
```
kdmccormick, line 119: It's not clear to me that a library is the "collection" of the context--I think that's mixing metaphors. Since we're making the assumption that a collection always belongs to a library, I would just be concrete and use the `lib_key` field that's already defined on your locator below. And, going along with my other request, I would spell it out as `library_key`.
```suggestion
    @property
    def library_key(self) -> LibraryLocatorV2:  # pragma: no cover
        """
        Get the key of the library that this collection belongs to.
        """
        raise NotImplementedError()
```

kdmccormick (changes requested):
Sorry for the late review. Hopefully these are easy to resolve. Other than these and Dave's point about having a slug, looks great.

opaque_keys/edx/keys.py
bradenmacdonald: Agreed that a collection is not a learning context and it should not subclass `LearningContextKey`. However, it can have a `library_key ` _field_ if that makes sense, since it does related to a library.

ChrisChV:
@kdmccormick Thanks for the review! I updated the PR with your suggestions

opaque_keys/edx/keys.py
```diff
@@ -89,6 +89,37 @@ def make_asset_key(self, asset_type: str, path: str) -> AssetKey:  # pragma: no
...
+    @property
+    @abstractmethod
+    def lib(self) -> str | None:  # pragma: no cover
+        """
+        The lib for this collection.
+        """
+        raise NotImplementedError()
+
+    @property
+    def library_key(self) -> LearningContextKey:  # pragma: no cover
```
ChrisChV, line 116: I can't use `LibraryLocatorV2` because of a circular import

pomegranited (approved):
@ChrisChV Could you do a version bump and add a changelog entry?
Otherwise this looks great 👍 

:+1:

- [x] I tested this while testing https://github.com/openedx/edx-platform/pull/35383
- [x] I read through the code
- [x] <strike>I checked for accessibility issues</strike> N/A
- [x] Includes documentation
- [x] <strike>User-facing strings are extracted for translation</strike> N/A

opaque_keys/edx/keys.py
bradenmacdonald: Can you put it in quotes? https://peps.python.org/pep-0484/#forward-references

opaque_keys/edx/locator.py
```diff
@@ -1620,3 +1621,59 @@ def html_id(self) -> str:
...
+        """
+        Construct a CollectionLocator
+        """
+        if not isinstance(lib_key, LibraryLocatorV2):
+            raise TypeError("lib_key must be a LibraryLocatorV2")
+
+        self._check_key_string_field("usage_id", usage_id, regexp=self.USAGE_ID_REGEXP)
+        super().__init__(
+            org=lib_key.org,
+            lib=lib_key.slug,
```
bradenmacdonald, line 1654: Do we need to store `org` and `lib` as separate fields? I would assume you can just story `lib_key` and create a `@property def org(self): return self.lib_key.org` accessor etc.

opaque_keys/edx/locator.py
```diff
@@ -1620,3 +1621,59 @@ def html_id(self) -> str:
...
+    """
+    When serialized, these keys look like:
+        lib-collection:org:lib:collection-id
+    """
+    CANONICAL_NAMESPACE = 'lib-collection'
+    KEY_FIELDS = ('org', 'lib', 'usage_id')
+    org: str
+    lib: str
+    lib_key: LibraryLocatorV2
+    usage_id: str
```
bradenmacdonald, line 1636: I'm not sure if `usage_id` is the best name for this slug field. `usage_id` usually refers to a whole `UsageKey`, not a short slug, which is usually called `block_id`. So this short slug is not a `usage_id`. I think it would be better called a `collection_id` or `slug` or `code`, but probably `collection_id`

opaque_keys/edx/locator.py
```diff
@@ -1620,3 +1621,59 @@ def html_id(self) -> str:
...
         return str(self)
+
+
+class LibraryCollectionLocator(CheckFieldMixin, LibraryCollectionKey):
+    """
+    When serialized, these keys look like:
+        lib-collection:org:lib:collection-id
+    """
+    CANONICAL_NAMESPACE = 'lib-collection'
+    KEY_FIELDS = ('org', 'lib', 'usage_id')
```
bradenmacdonald, line 1632: Based on [this example](https://github.com/openedx/opaque-keys/blob/e3ff649fb59dfbc316fb99f33e3fed57e36f8201/opaque_keys/edx/locator.py#L649) I think this should be:

```suggestion
    KEY_FIELDS = ('lib_key', 'usage_id')
```

Or `('lib_key', 'collection_id')` - see below.

opaque_keys/edx/keys.py
ChrisChV: The quotes don't work, but I found this [solution](https://github.com/openedx/opaque-keys/pull/335/commits/168c373f4adafe7c81f707baef3a971f7404acd7#diff-6117aaf0e3d3aaf8569961ad62a10f24b38af67ea69267477abacaf7f071676dR15)

opaque_keys/edx/locator.py
ChrisChV: Updated: https://github.com/openedx/opaque-keys/pull/335/commits/168c373f4adafe7c81f707baef3a971f7404acd7

opaque_keys/edx/locator.py
ChrisChV: Updated: https://github.com/openedx/opaque-keys/pull/335/commits/168c373f4adafe7c81f707baef3a971f7404acd7

opaque_keys/edx/locator.py
ChrisChV: Updated: https://github.com/openedx/opaque-keys/pull/335/commits/168c373f4adafe7c81f707baef3a971f7404acd7

opaque_keys/edx/keys.py
kdmccormick: @bradenmacdonald Sure, having a library_key field makes sense.

@ChrisChV Is there any reason we need both a field `LibraryCollectionLocator.lib_key` and an identical property `LibraryCollectionLocator.library_key` ? Could there just be a field named `LibraryCollectionLocator.library_key` ? Or does that play poorly with inheritance?

bradenmacdonald:
I think we just need to address @kdmccormick 's question about combining `LibraryCollectionLocator.lib_key` and  `LibraryCollectionLocator.library_key`, and then this is good to go from my perspective.

Please also include a [version bump](https://github.com/openedx/opaque-keys/blob/e133f28f39ec2b059c037fcbcc1fbfac380f8af2/opaque_keys/__init__.py#L17) so it'll create a new package version.

opaque_keys/edx/keys.py
ChrisChV: I combined both fields here: https://github.com/openedx/opaque-keys/pull/335/commits/ce740812b6874493989bdd1c6502378911c51e91

ChrisChV:
> Please also include a [version bump](https://github.com/openedx/opaque-keys/blob/e133f28f39ec2b059c037fcbcc1fbfac380f8af2/opaque_keys/__init__.py#L17) so it'll create a new package version.

@bradenmacdonald Done!

bradenmacdonald (approved):

opaque_keys/edx/locator.py
```diff
@@ -1620,3 +1621,58 @@ def html_id(self) -> str:
...
+    CANONICAL_NAMESPACE = 'lib-collection'
+    KEY_FIELDS = ('library_key', 'collection_id')
+    library_key: LibraryLocatorV2
+    collection_id: str
+
+    __slots__ = KEY_FIELDS
+    CHECKED_INIT = False
+
+    # Allow usage IDs to contian unicode characters
+    USAGE_ID_REGEXP = re.compile(r'^[\w\-.]+$', flags=re.UNICODE)
```
kdmccormick, line 1640:
```suggestion
    COLLECTION_ID_REGEXP = re.compile(r'^[\w\-.]+$', flags=re.UNICODE)
```

opaque_keys/edx/locator.py
```diff
@@ -1620,3 +1621,58 @@ def html_id(self) -> str:
...
+
+    def __init__(self, library_key: LibraryLocatorV2, collection_id: str):
+        """
+        Construct a CollectionLocator
+        """
+        if not isinstance(library_key, LibraryLocatorV2):
+            raise TypeError("library_key must be a LibraryLocatorV2")
+
+        self._check_key_string_field("collection_id", collection_id, regexp=self.USAGE_ID_REGEXP)
```
kdmccormick, line 1649:
```suggestion
        self._check_key_string_field("collection_id", collection_id, regexp=self.COLLECTION_ID_REGEXP)
```

kdmccormick (approved):
One change request, but LGTM otherwise!

opaque_keys/edx/locator.py
ChrisChV: Thanks! Updated here: https://github.com/openedx/opaque-keys/pull/335/commits/780e4b264d393edc7e15dd3a250ec80d7f19d025

opaque_keys/edx/locator.py
ChrisChV: Updated here: https://github.com/openedx/opaque-keys/pull/335/commits/780e4b264d393edc7e15dd3a250ec80d7f19d025

ChrisChV:
@bradenmacdonald It's ready for merge

openedx-webhooks:
@ChrisChV 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/opaque-keys&entry.867055334=https://github.com/openedx/opaque-keys/pull/335&entry.1484655318=https://github.com/ChrisChV&entry.752974735=2024-08-27+01:51&entry.1917517419=2024-09-04+17:06&entry.2133058324=Yes) so we can improve your experience in the future.

ChrisChV:
@bradenmacdonald I see the [tag](https://github.com/openedx/opaque-keys/releases/tag/2.11.0) for this version, but the pip package is not updated

bradenmacdonald:
@ChrisChV It seems like I needed to [publish a "Release"](https://github.com/openedx/opaque-keys/releases/tag/2.11.0). I've done that now so you should see it on PyPI soon.

pomegranited:
Hi @kdmccormick  Sorry for pinging you directly, but we're having trouble bumping this requirement in edx-platform, cf https://github.com/openedx/edx-platform/pull/35383. It doesn't seem to be actually constrained in anyone else's `.in` files that we can find, but it is in their `.txt` files.

What's the procedure for updating these fundamental packages?

PR #350, opened 2024-11-18
fix: register CourseLocator adapter for postgresql

qasimgulzar:
This PR is subsequent of [#35762](https://github.com/openedx/edx-platform/pull/35762)


qasimgulzar:
@ormsbee I just pushed couple of new changes to add support for postgresql, please give it a look when you get sometime.

qasimgulzar:
I have moved this change inside edx-platform repo. As it seems more appropriate place to register adapters.

PR #369, opened 2025-03-13
feat: Create `LibraryContainerLocator` for Library Containers [FC-0083]

ChrisChV:
## Description

- Refactor `LibraryCollectionKey` to `LibraryItemKey` to use it as the general class for `LibraryCollectionLocator` and `LibraryContainerLocator`.
- Creation of `LibraryContainerLocator`
- Add `LibraryUsageLocatorV2` also as `LibraryItemKey`

## Support Information

- Github issues:
    - https://github.com/openedx/openedx-learning/issues/282
    - https://github.com/openedx/frontend-app-authoring/issues/1704
- Used in: https://github.com/openedx/edx-platform/pull/36379
- Internal ticket: [FAL-4108](https://tasks.opencraft.com/browse/FAL-4108)

## Testing instructions

- Check all the code.
- Verify that the test covers the changes.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,6 @@
 # 2.12.0
 
-* Refactor: Rename LibraryCollectionKey to LibraryElementKey.
+* Refactor: Rename LibraryCollectionKey to LibraryItemKey.
 * Added LibraryContainerLocator.
 
 # 2.11.0
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,3 +1,8 @@
+# 2.12.0
+
+* Refactor: Rename LibraryCollectionKey to LibraryElementKey.
+* Added LibraryContainerLocator.
+
 # 2.11.0
 
 * Added LibraryCollectionKey and LibraryCollectionLocator
--- a/opaque_keys/edx/keys.py
+++ b/opaque_keys/edx/keys.py
@@ -93,10 +93,9 @@ class CourseKey(LearningContextKey):
         raise NotImplementedError()
 
 
-class LibraryElementKey(OpaqueKey):
+class LibraryItemKey(OpaqueKey):
     """
-    An :class:`opaque_keys.OpaqueKey` identifying a particular element in a library
-    that is not an Xblock.
+    An :class:`opaque_keys.OpaqueKey` identifying a particular item in a library.
     """
     KEY_TYPE = 'library_element_key'
     library_key: LibraryLocatorV2
--- a/opaque_keys/edx/keys.py
+++ b/opaque_keys/edx/keys.py
@@ -103,11 +103,12 @@ class LibraryElementKey(OpaqueKey):
     __slots__ = ()
 
     @property
+    @abstractmethod
     def org(self) -> str | None:  # pragma: no cover
         """
         The organization that this object belongs to.
         """
-        return self.library_key.org
+        raise NotImplementedError()
 
 
 class DefinitionKey(OpaqueKey):
--- a/opaque_keys/edx/keys.py
+++ b/opaque_keys/edx/keys.py
@@ -97,7 +97,7 @@ class LibraryItemKey(OpaqueKey):
     """
     An :class:`opaque_keys.OpaqueKey` identifying a particular item in a library.
     """
-    KEY_TYPE = 'library_element_key'
+    KEY_TYPE = 'library_item_key'
     library_key: LibraryLocatorV2
     __slots__ = ()
 
--- a/opaque_keys/edx/keys.py
+++ b/opaque_keys/edx/keys.py
@@ -93,21 +93,21 @@ class CourseKey(LearningContextKey):
         raise NotImplementedError()
 
 
-class LibraryCollectionKey(OpaqueKey):
+class LibraryElementKey(OpaqueKey):
     """
-    An :class:`opaque_keys.OpaqueKey` identifying a particular Library Collection object.
+    An :class:`opaque_keys.OpaqueKey` identifying a particular element in a library
+    that is not an Xblock.
     """
-    KEY_TYPE = 'collection_key'
+    KEY_TYPE = 'library_element_key'
     library_key: LibraryLocatorV2
-    collection_id: str
     __slots__ = ()
 
     @property
     def org(self) -> str | None:  # pragma: no cover
         """
-        The organization that this collection belongs to.
+        The organization that this object belongs to.
         """
-        raise NotImplementedError()
+        return self.library_key.org
 
 
 class DefinitionKey(OpaqueKey):
--- a/opaque_keys/edx/locator.py
+++ b/opaque_keys/edx/locator.py
@@ -16,7 +16,7 @@ from typing_extensions import Self  # For python 3.11 plus, can just use "from t
 
 from opaque_keys import OpaqueKey, InvalidKeyError
 from opaque_keys.edx.keys import AssetKey, CourseKey, DefinitionKey, \
-    LearningContextKey, UsageKey, UsageKeyV2, LibraryElementKey
+    LearningContextKey, UsageKey, UsageKeyV2, LibraryItemKey
 
 log = logging.getLogger(__name__)
 
@@ -1623,7 +1623,7 @@ class LibraryUsageLocatorV2(CheckFieldMixin, UsageKeyV2):
         return str(self)
 
 
-class LibraryCollectionLocator(CheckFieldMixin, LibraryElementKey):
+class LibraryCollectionLocator(CheckFieldMixin, LibraryItemKey):
     """
     When serialized, these keys look like:
         lib-collection:org:lib:collection-id
@@ -1654,7 +1654,7 @@ class LibraryCollectionLocator(CheckFieldMixin, LibraryElementKey):
     @property
     def org(self) -> str | None:  # pragma: no cover
         """
-        The organization that this object belongs to.
+        The organization that this Collection belongs to.
         """
         return self.library_key.org
 
@@ -1677,7 +1677,7 @@ class LibraryCollectionLocator(CheckFieldMixin, LibraryElementKey):
             raise InvalidKeyError(cls, serialized) from error
 
 
-class LibraryContainerLocator(CheckFieldMixin, LibraryElementKey):
+class LibraryContainerLocator(CheckFieldMixin, LibraryItemKey):
     """
     When serialized, these keys look like:
         lct:org:lib:ct-type:ct-id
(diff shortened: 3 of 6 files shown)


openedx-webhooks:
Thanks for the pull request, @ChrisChV!


This repository is currently maintained by `@openedx/axim-engineering`.


Once you've gone through the following steps feel free to tag them in a comment and let them know that your changes are ready for engineering review.

<details><summary>:radio_button: Get product approval</summary>

</details>

<details><summary>:radio_button: Provide context</summary>

- Dependencies
- Blockers
- Timeline information
- Partner information
- Supporting documentation
</details>





<details><summary>:radio_button: Get a green build</summary>

</details><details>


---

<details><summary>Where can I find more information?</summary>

</details>

<details><summary>When can I expect my changes to be merged?</summary>

- The need for product review

</details>

bradenmacdonald:
Is it possible for `LibraryUsageLocatorV2` to also be a LibraryElementkey, or can each locator only be a single type of key? (I guess it's more important that it be a UsageKey if it can only be one).

CHANGELOG.rst
```diff
@@ -1,3 +1,8 @@
+# 2.12.0
+
+* Refactor: Rename LibraryCollectionKey to LibraryElementKey.
```
bradenmacdonald, line 3: Nit: we have been using the term "item" in a few place to refer to any thing in a library (block/collection/container) so I would slightly prefer that over "element"

navinkarkera (approved):
@ChrisChV Looks good.
:+1:
- [ ] I tested this: Verified tests.
- [x] I read through the code
- [x] I checked for accessibility issues
- [x] Includes documentation

opaque_keys/edx/locator.py
```diff
@@ -1676,3 +1675,65 @@ def _from_string(cls, serialized: str) -> Self:
...
         except (ValueError, TypeError) as error:
             raise InvalidKeyError(cls, serialized) from error
+
+
+class LibraryContainerLocator(CheckFieldMixin, LibraryElementKey):
+    """
+    When serialized, these keys look like:
+        lct:org:lib:ct-type:ct-id
+    """
+    CANONICAL_NAMESPACE = 'lct'  # "Library Container"
```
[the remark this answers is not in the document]
ChrisChV: @navinkarkera Yes, other larger possibilities were ruled out; you can see that conversation here: https://github.com/openedx/openedx-learning/issues/282#issuecomment-2715951009

CHANGELOG.rst
ChrisChV: Updated here https://github.com/openedx/opaque-keys/pull/369/commits/fc0cff1bc932993f47f6fc0b58bb3cd882a1ca31

ChrisChV:
> Is it possible for LibraryUsageLocatorV2 to also be a LibraryElementkey, or can each locator only be a single type of key? 

@bradenmacdonald  I've tried this, and it causes conflicts with the [KEY_TYPE](https://github.com/openedx/opaque-keys/pull/369/files#diff-6117aaf0e3d3aaf8569961ad62a10f24b38af67ea69267477abacaf7f071676dR100) (it inherits two different values ​​for this property). Apparently, a locator can only have one `KEY_TYPE`.

ChrisChV:
@bradenmacdonald It's ready for your review

bradenmacdonald (approved):
Looks good! I tested this out in https://github.com/openedx/edx-platform/pull/36371 though I haven't tried it with the latest rename from Element -> Item

pomegranited:
Hi @ormsbee or @kdmccormick  -- can you help get this merged and a new release created?

opaque_keys/edx/keys.py
```diff
@@ -93,19 +93,19 @@ def make_asset_key(self, asset_type: str, path: str) -> AssetKey:  # pragma: no
...
 
 
-class LibraryCollectionKey(OpaqueKey):
+class LibraryItemKey(OpaqueKey):
     """
-    An :class:`opaque_keys.OpaqueKey` identifying a particular Library Collection object.
+    An :class:`opaque_keys.OpaqueKey` identifying a particular item in a library.
     """
-    KEY_TYPE = 'collection_key'
+    KEY_TYPE = 'library_element_key'
```
pomegranited, line 100: If we're renaming this to `LibraryItemKey` instead of `LibraryElementKey`, shouldn't this key type be `library_item_key`?

opaque_keys/edx/tests/test_container_locators.py
```diff
@@ -0,0 +1,75 @@
...
+        with self.assertRaises(ValueError):
+            LibraryContainerLocator(library_key=library_key, container_type='unit', container_id='usage-!@#{$%^&*}')
+
+    def test_key_constructor_bad_type(self):
+        library_key = LibraryLocatorV2(org="TestX", slug="lib1")
+
+        with self.assertRaises(ValueError):
+            LibraryContainerLocator(library_key=library_key, container_type='unit-!@#{$%^&*}', container_id='usage')
+
+    def test_key_from_string(self):
```
pomegranited, line 62: @ChrisChV Could you please add a test for converting a `lib-collection:...` string to a `LibraryItemKey` key so we can ensure backwards compatibility? (I'm asking because I think there's a bug with this somewhere too).

opaque_keys/edx/keys.py
ChrisChV: You are right. Updated: https://github.com/openedx/opaque-keys/pull/369/commits/75ecbe6e610e73e6d4def545106e5fb89356840e

kdmccormick (changes requested):
Don't we need LibraryContainerLocators to be UsageKeys so that we can store xblock state against units, sections, and subsections?

ChrisChV:
> Don't we need LibraryContainerLocators to be UsageKeys so that we can store xblock state against units, sections, and subsections?

Hi @kdmccormick, could you give me more context about this? What is the xblock state that we need to store in the containers?

kdmccormick:
Sure @ChrisChV . Honestly, I actually cannot think of a reason to store state against containers in libraries right now. Still, I think that this would be good future-proofing, and it is fairly harmless if we don't end up needing it. Here are some of the ways I think it could pay off:

* It is not clear yet how courses will be represented in Learning Core. One idea is that courses will be stored _within libraries_. That would mean...
  * we would store a student's [`position`](https://github.com/openedx/edx-platform/blob/master/xmodule/seq_block.py#L75) as user state against an `lct:`. key. 
  * furthermore, we will need to store student state for various selectors (e.g. [`ItemBank.selected`](https://github.com/openedx/edx-platform/blob/f20b9cf466083b08e67d1a6ee2023b38e10f385b/xmodule/item_bank_block.py#L72)); I'm not sure if that state will get stored against the selector or its container, but if the latter, then we that will be state stored against `lct:`.
  * Many many edx-platform Python APIs and HTTP APIs a expect UsageKeys as a reference to units and sequences. Examples include the CourseBlocks (block transformers) and SequenceMetadata APIs.
* We know that we want to move away from the idea that "containers are XBlocks", but we are several years away from that. In the meantime, I believe that the Learning Core runtime will need to be able to return containers as if they are XBlocks, and that is only possible if they have usage keys.

kdmccormick:
@ormsbee ^ I could use a sanity check from you on this, if you have time.

bradenmacdonald:
Jumping in with some quick thoughts on this: I'm not super opposed to making these UsageKeys, but it doesn't make sense to me and I think it's cleaner not to do so.

> One idea is that courses will be stored within libraries. That would mean... 

If you mean courses will be stored within libraries and learners will learn from those courses within libraries, I haven't heard about that and don't like it in principle. I do see how having a course within a library is arguably a natural extension of having units, subsections, and sections - but learners wouldn't be accessing any of those things within the library generally; only when they're copied into a course. "True" independent courses might be in the same learning package, but I don't think they'd be accessed directly by learners within a library. What's more, courses aren't really "containers" and generally have their own keys `course-v1` etc.

> furthermore, we will need to store student state for various selectors (e.g. [ItemBank.selected](https://github.com/openedx/edx-platform/blob/f20b9cf466083b08e67d1a6ee2023b38e10f385b/xmodule/item_bank_block.py#L72));

That's only required for learning in courses. We don't support `ItemBank` in libraries at all, because it's an XBlock with children, and the runtime doesn't support it. Eventually we will implement ItemBank-like functionality on top of containers (and "selectors") and then it will be allowed in libraries and support student state, but it wouldn't be XBlock student state. It will be a new "container"/"outline" student state, and would work fine with `lct:` keys that are not `UsageKeys`.

> Many many edx-platform Python APIs and HTTP APIs a expect UsageKeys as a reference to units and sequences. Examples include the CourseBlocks (block transformers) and SequenceMetadata APIs.

None of those are APIs that we use with content libraries, and if they were, then they'd need to be updated to support non-XBlock containers anyways, since that's what we want to move toward. I see this (having `lct:` as a non-UsageKey) as a forcing function to help us aim toward the new architecture (containers aren't XBlocks).

> the Learning Core runtime will need to be able to return containers as if they are XBlocks, and that is only possible if they have usage keys.

I don't think we need that (in libraries)? They aren't implemented as XBlocks, and can't be used as XBlocks in the Library / Learning Core runtime. That runtime doesn't support XBlocks with children so it can't support Units as XBlocks.

ormsbee:
I'll try to write more tomorrow, but my quick take on this is that I don't think this needs to be a UsageKey. Also, I don't want to get in the mindset that we need a parallel set of container UsageKeys because of where the data is being stored. It's true that at some point we're going to be serving container content data out of Learning Core, if nothing else when we move course content to it. But we're going to need some mapping of existing block-v1: style UsageKeys to the corresponding course content in a LearningPackage anyhow, for backwards compatibility. That mapping doesn't even have to be 1:1 (e.g. the CCX use case).

I am fuzzy on exactly how we transition from a world where it's XBlocks all the way down to one where we shove the containers out of the XBlock runtime. The Unit seems like a great place to start because it has no user state for the most part (ignoring custom asides and the fact that technically it has a position field that I'm almost sure is never actually used). I expect that working out how to sequence that transition will be one of the thorniest pieces of this whole affair.

kdmccormick (approved):
OK, that all makes sense. Then let's merge this 🚀

ChrisChV:
> OK, that all makes sense. Then let's merge this 🚀

Thanks! :rocket: :rocket: @kdmccormick @ormsbee could you help me merge this and create a new release?

kdmccormick:
@ormsbee any last-minute objections?

I still think these might be UsageKeys, but we can retroactively add that base class in the future if it turns out to be true.

ormsbee:
No objections. Agreed that we can add it in the future if we need to.

rpenido:
> Can you merge @ChrisChV or should I?

He doesn't have merge rights on this repo, so it would be great if you could merge it.

kdmccormick:
Released: https://github.com/openedx/opaque-keys/releases/tag/2.12.0

PR #434, opened 2026-03-25
feat!: rename LibraryCollectionLocator.collection_id to collection_code

kdmccormick:
BREAKING CHANGE: `LibraryCollectionLocator.collection_id` is renamed to `LibraryCollectionLocator.collection_code`.  There is a readonly `collection_id`` property for backwards compatibility, but calls to the constructor which use the `collection_id=` kwarg must be updated to `collection_code=`.

Part of: https://github.com/openedx/openedx-core/issues/322

Related openedx-core PR: https://github.com/openedx/openedx-core/pull/504
Related openedx-platform PR: https://github.com/openedx/openedx-platform/pull/38214

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,3 +1,9 @@
+# 4.0.0
+
+* Breaking: ``LibraryCollectionLocator.collection_id`` is renamed to ``LibraryCollectionLocator.collection_code``.
+  There is a readonly ``collection_id`` property for backwards compatibility, but calls to the constructor
+  which use the ``collection_id=`` kwarg must be updated to ``collection_code=``.
+
 # 3.1.0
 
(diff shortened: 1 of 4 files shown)


bradenmacdonald (commented):
This looks good, but there are a lot of other similar things I'd like to fix. Is your idea to do them all incrementally, releasing many breaking versions of opaque-keys ?

`LibraryContainerLocator.container_type` -> `container_type_code`
`LibraryContainerLocator.container_id` -> `container_code`
`LibraryUsageLocatorV2.usage_id` -> `usage_code`
`LibraryLocatorV2.org` -> `org_code`
`LibraryLocatorV2.slug` -> `library_code`

and that's just for V2 libraries.

PR #461, opened 2026-07-27
feat: modernize to uv + pyproject.toml + semantic-release

irfanuddinahmad:
## Summary

Modernize `opaque-keys` to `uv` + `pyproject.toml` (PEP 621/735) + `python-semantic-release`, and adopt `src/` layout.

Part of openedx/public-engineering#506 (tracked in openedx/public-engineering#516).

- Move `opaque_keys/` to `src/opaque_keys/` (see "`src/` layout" below).
- Replace `setup.py` with `pyproject.toml` (PEP 621 static metadata); remove the
  hardcoded `__version__` in favor of `importlib.metadata` + `setuptools-scm`.
- Switch from pip-compile to `uv` with PEP 735 dependency groups (`test-base`,
  `test`, `django42`, `quality`, `doc`, `ci`, `dev`); commit `uv.lock`.
- Update `tox.ini` to use `tox-uv`'s `uv-venv-lock-runner`.
- Update CI to install `uv` (pinned to a commit SHA) and run `uv sync` / `uv run tox`.
- Add `python-semantic-release` + `release.yml`, publishing to PyPI via OIDC.
  `commitlint.yml` already existed and needed no changes.

## `src/` layout

Per the 2026-07-15 decision on openedx/public-engineering#506 ("src/ layout is in
scope for this cycle"), motivated by flat-layout editable installs of packages like
opaque-keys not being reliably resolvable by `mypy`. I verified the cited precedent
directly rather than trusting the decision comment's own description of it:
`openedx/xblocks-extra` does have a real `src/` directory (used as the reference for
the `package-dir`/`packages.find` config here); `openedx/XBlock`, despite being named
as the other precedent, does **not** actually have one as of this writing.

`mypy.ini` switched from `files = opaque_keys` (a literal path, now stale) to
`packages = opaque_keys` (import-based resolution via the editable install) — that's
the actual fix for the problem this decision cites, not just a path rename. Verified
locally: editable install resolves `opaque_keys` to `src/opaque_keys/__init__.py`,
`mypy`/`pylint`/`pycodestyle` all pass, the full test suite passes (see Testing
Notes), and a built wheel correctly installs `opaque_keys` at the top level.

One side effect: the top-level `settings` test-settings package (previously bundled
in the wheel only as an artifact of `setup.py`'s bare `find_packages()`) is no longer
picked up, since it lives outside `src/`. That's now the correct scope for it.

## Removed

**Deleted files:** `setup.py`, `requirements/*.in`, `requirements/*.txt`,
`.github/workflows/pypi-publish.yml` (superseded by `release.yml`'s
`publish_to_pypi` job — the old token-based workflow triggered on
`release: published` and would otherwise race the new OIDC-based publish once
semantic-release starts creating releases/tags).

## Not included

- **Ruff**: per the 2026-07-15 discussion on openedx/public-engineering#506, ruff
  adoption is out of scope for this migration cycle (its own separate epic). Existing
  pylint/pycodestyle/mypy config is left untouched.
- **`edx-platform`-style constraint-sync script**: not applicable — this repo never
  had the `requirements/edx-platform-constraints.txt` + `check_pins.py` mechanism
  that pattern replaces.

## Known out-of-scope gaps (flagging, not silently working around)

- **PyPI trusted publisher (OIDC) — pre-merge blocker.** `release.yml`'s
  `publish_to_pypi` job publishes via OIDC (`id-token: write`, no stored
  credentials) instead of the old `PYPI_UPLOAD_TOKEN`. This requires a PyPI
  project maintainer to add a trusted publisher for `edx-opaque-keys` (PyPI project
  → **Publishing** → **Add a new publisher** → GitHub, repo `openedx/opaque-keys`,
  workflow `release.yml`, no environment) before the *first* release can actually
  publish. Not a blocker for reviewing/merging this PR, but `publish_to_pypi` will
  fail on the first push to `master` until it's done.
- **`.github/workflows/upgrade-python-requirements.yml`.** This repo has one,
  calling `openedx/.github`'s shared reusable workflow, which hardcodes
  `ADD_PATHS="requirements"` for its PR-creation step with no `add_paths`-style
  input exposed via `workflow_call`. Now that `requirements/` is deleted, `make
  upgrade`'s real output (`uv.lock`/`pyproject.toml` changes) won't match that
  hardcoded glob — the scheduled job will keep running but silently stop producing
  real dependency-upgrade PRs (empty diff, no failure, no alert). This can't be
  fixed from this repo; it needs a fix in `openedx/.github` itself (parameterize
  `add_paths`, or auto-detect `uv.lock` vs. `requirements/`).

## Versioning

Dynamic, via `setuptools-scm` reading the latest git tag. The latest tag (`4.0.0`)
matches the latest version published on PyPI, so `python-semantic-release` will
compute the next version from that baseline correctly.

## Testing Notes

Verified locally (macOS, Python 3.12), both before and after the `src/` layout move:
- `uv lock` resolves cleanly (86 packages); `uv lock --check` confirms it stays
  in sync with `pyproject.toml` after each change.
- `uv sync --group dev` succeeds; editable install resolves `opaque_keys` to
  `src/opaque_keys/__init__.py`.
- `uv run tox -e django42` — 407 passed.
- `uv run tox -e without-django` — 394 passed.
- `uv run tox -e docs` — builds cleanly.
- `uv run tox -e quality` — `mypy` passes cleanly (30 source files) using
  `packages = opaque_keys`; `pylint`/`pycodestyle` pass (targeting `src/opaque_keys`
  now). `pylint` initially reported `no-member` false positives on Django field
  `super()` calls once the quality env correctly started installing Django (it
  silently hadn't before, under the old pip-based tooling) — fixed with targeted
  `# pylint: disable=no-member` comments; confirmed not a real bug.
- `python -m build`, including with `SETUPTOOLS_SCM_PRETEND_VERSION` set (the
  mechanism `python-semantic-release`'s `build_command` relies on) — both build
  correctly-versioned wheels/sdists with the expected metadata, entry points, and
  (post-`src/`-move) top-level `opaque_keys` package layout.
- CI-only issues found and fixed after opening this PR: `codecov/patch` (an
  unreachable `importlib.metadata` fallback line, fixed with `pragma: no cover`,
  and fully moving coverage config off a stale `.coveragerc` that was still
  shadowing `pyproject.toml`); `codecov/project` (a one-time ~1.6 point drop from
  94.03%→92.46% once test files were correctly excluded from coverage measurement,
  not a real regression — patch coverage is 100% — absorbed with a `codecov.yml`
  project threshold, matching the same pattern applied to `ccx-keys`/
  `openedx-filters` in this same effort); `.readthedocs.yaml` (still pointed at
  the deleted `requirements/doc.txt`, switched to `method: uv`/`groups: [doc]`,
  matching `openedx/XBlock`'s working config).

## Code reviewer notes

- `django42` is the only Django version group for now (this repo's `tox.ini`
  envlist has only ever tested one Django version at a time); adding a
  `django52`-style group later just needs `[tool.uv].conflicts` if two Django
  pins need to coexist in one lockfile.

---

_This PR was created with Claude Code._


openedx-webhooks:
Thanks for the pull request, @irfanuddinahmad!


This repository is currently maintained by `@openedx/axim-engineering`.


Once you've gone through the following steps feel free to tag them in a comment and let them know that your changes are ready for engineering review.

<details><summary>:radio_button: Get product approval</summary>

</details>

<details><summary>:radio_button: Provide context</summary>

- Dependencies
- Blockers
- Timeline information
- Partner information
- Supporting documentation
</details>





<details><summary>:radio_button: Get a green build</summary>

</details><details>


---

<details><summary>Where can I find more information?</summary>

</details>

<details><summary>When can I expect my changes to be merged?</summary>

- The need for product review

</details>

mphilbrick211:
Hi @irfanuddinahmad - flagging that some checks and branch conflicts have popped up.

irfanuddinahmad:
Rebased on master — conflicts resolved, CI green.

irfanuddinahmad:
3 required checks are stuck on "Expected — Waiting for status to be reported" and won't ever resolve: `Tests (ubuntu-latest, 3.12, django42)`, `Tests (ubuntu-latest, 3.12, docs)`, `Tests (ubuntu-latest, 3.12, quality)`. Branch protection requires them under GitHub's old auto-generated matrix-name format, but `ci.yml`'s job now reports under `name: ${{ matrix.toxenv }}` (bare `django42`/`docs`/`quality`) -- the same convention already used successfully in `xblocks-extra#77`, `xblocks-core#297`, and `forum#290`, all clean/merged once their branch protection was updated to match. Could someone with admin access update this repo's required-status-checks list the same way?

.github/workflows/ci.yml
```diff
@@ -1,39 +1,45 @@
 name: Python CI
 
 on:
-  push:
```
salman2013, line 4: Please put it back as per
https://github.com/openedx/crowdsourcehinter/pull/283/changes#r3753071643

.github/workflows/ci.yml
```diff
@@ -1,39 +1,45 @@
...
 
 jobs:
   run_tests:
-    name: Tests
+    name: ${{ matrix.toxenv }}
     runs-on: ${{ matrix.os }}
+    permissions:
+      contents: read
     strategy:
+      fail-fast: false
```
salman2013: Any specific reason to make fail-fast: false

codecov.yml
```diff
@@ -0,0 +1,11 @@
+coverage:
```
salman2013: We do not have to add this file.

pyproject.toml
```diff
@@ -0,0 +1,189 @@
...
+
+[project]
+name = "edx-opaque-keys"
+dynamic = ["version"]
+description = "Create and introspect course and xblock identities through a clear and consistent API"
+readme = "README.rst"
+license = "AGPL-3.0"
+license-files = ["LICENSE"]
+authors = [
+    {name = "edX", email = "oscm@edx.org"},
```
salman2013: This is not correct as per following information

https://github.com/openedx/opaque-keys/blob/6009a24a57b5c549aad953109b6f973d99e2acbe/.github/workflows/upgrade-python-requirements.yml#L21
https://github.com/openedx/opaque-keys/blob/6009a24a57b5c549aad953109b6f973d99e2acbe/catalog-info.yaml#L18

pyproject.toml
```diff
@@ -0,0 +1,189 @@
...
+    "elasticsearch<7.14.0",
+]
+[tool.edx_lint]
+uv_constraints = []
+
+[tool.semantic_release]
+build_command = "python -m pip install --upgrade build && SETUPTOOLS_SCM_PRETEND_VERSION=$NEW_VERSION python -m build"
+major_on_zero = false
+allow_zero_version = true
+tag_format = "{version}"
```
salman2013: I think this need to be remove as semantic release take versions format like "v(x.y.z)"

salman2013 (changes requested):

.github/workflows/ci.yml
irfanuddinahmad: This matches sample-plugin's current pattern — ci.yml only runs on pull_request + workflow_call, no direct push. release.yml already triggers on push: branches: [master] and calls ci.yml via workflow_call, so CI still runs on every push to master — just without triggering twice. Assuming that behavior holds (worth double-checking a run on master once merged), we'd rather not reintroduce the duplicate trigger.

.github/workflows/ci.yml
irfanuddinahmad: Fair catch, no real reason for it — removed, matching sample-plugin's default behavior.

codecov.yml
irfanuddinahmad: Checked — sample-plugin's backend-plugin-sample/codecov.yml has the same shape (project target auto, patch target 100%). Given that precedent, keeping this rather than letting the one-time coverage-scope-change fail CI.

pyproject.toml
irfanuddinahmad: sample-plugin uses {name = "Open edX Project", email = "oscm@openedx.org"} (not edX/oscm@edx.org, which is what XBlock still has). Updated to match sample-plugin's convention.

pyproject.toml
irfanuddinahmad: sample-plugin doesn't set tag_format (defaults to v{version}, confirmed via its actual tags: v3.8.0, etc.). But opaque-keys already has 20+ releases tagged without the v prefix (4.0.0, 3.1.0, ...) matching what's on PyPI — switching to the default here would break semantic-release's ability to recognize those as prior releases. Keeping tag_format = "{version}" to preserve that continuity.

.github/workflows/ci.yml
irfanuddinahmad: Per code owners: restored, and the double-run on master pushes is accepted as fine.

pyproject.toml
irfanuddinahmad: Update: switching to tag_format = "v{version}" — planning a manual v-prefixed release to bridge the existing unprefixed tag history (4.0.0, 3.1.0, ...) so future releases stay consistent going forward.

codecov.yml
irfanuddinahmad: Update: removed codecov.yml. On reflection it didn't actually change any pass/fail outcome (project's target: auto with no threshold buffer matches Codecov's own default, and patch was already at 100%, above what it set) and isn't required by this migration itself -- keeping the touch minimal.

salman2013 (approved):

pyproject.toml
```diff
@@ -0,0 +1,189 @@
...
+    "tox-uv",
+]
+dev = [
+    {include-group = "quality"},
+    {include-group = "ci"},
+    {include-group = "doc"},
+]
+
+[tool.uv]
+package = true
```
feanil: `dev` includes `quality`, `ci` and `doc`, and uv defaults `default-groups` to `["dev"]`, so `uv sync --group ci` in `ci.yml` installs Sphinx and the rest of the doc tree on all four matrix legs.

```
uv sync --group ci                      -> 88 packages
uv sync --no-default-groups --group ci  -> 63 packages
```

Add `default-groups = []` here.  `package = true` is a different setting and doesn't cover this.

pyproject.toml
```diff
@@ -0,0 +1,189 @@
+[build-system]
+requires = ["setuptools>=64", "setuptools-scm>=8.0"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "edx-opaque-keys"
+dynamic = ["version"]
+description = "Create and introspect course and xblock identities through a clear and consistent API"
+readme = "README.rst"
+license = "AGPL-3.0"
```
feanil: `AGPL-3.0` is deprecated in the SPDX list, and under `[project].license` this is now published as a validated expression rather than as free text.  The built wheel's METADATA carries:

```
License-Expression: AGPL-3.0
```

`LICENSE:637-638` grants "either version 3 of the License, or (at your option) any later version", so this should be `AGPL-3.0-or-later`.

.github/workflows/release.yml
```diff
@@ -0,0 +1,90 @@
...
+      with:
+        ref: ${{ github.ref_name }}
+
+    - run: git reset --hard ${{ github.sha }}
+
+    - name: Python Semantic Release
+      id: release
+      uses: python-semantic-release/python-semantic-release@9a026e9303981c866c3425723009becb2437c757 # v10.6.2
+      with:
+        github_token: ${{ secrets.OPENEDX_SEMANTIC_RELEASE_GITHUB_TOKEN }}
```
feanil: Use `secrets.GITHUB_TOKEN` here.  The org secret isn't needed going forward, and sample-plugin, DoneXBlock, xblocks-core and xblocks-extra are all on the default token already.

